PR #7, opened 2017-03-14
Metal Quartz Icon: #6 add options to render Glyphicons and Lexicon Font Awesome Icons

pat270:


jbalsas:

packages/metal-quartz-icon/demos/index.html
```diff
@@ -9,32 +9,57 @@
...
 <body>
 	<h1>Metal Quartz Icon</h1>
-
 	<div id="quartzIcons"></div>
 
+	<h1>Glyphicons</h1>
+	<div id="glyphicons"></div>
+
+	<h1>Lexicon Fa</h1>
+	<div id="lexiconFa"></div>
```
jbalsas: Should this be "Quartz Fa" and "quartzFa"? I know internally we're still going to be using the `lexicon-fa` class for now, but just to remove confusion as much as possible :)

packages/metal-quartz-icon/src/MetalQuartzIcon.js
```diff
@@ -9,6 +9,8 @@ import templates from './MetalQuartzIcon.soy.js';
 /**
  * Implementation of the Metal Quartz Icon.
  * @see  {@link http://liferay.github.io/lexicon/content/icons-lexicon/|Lexicon Definition}
+ * {@link http://liferay.github.io/lexicon/content/icons-font-awesome/}
+ * {@link http://liferay.github.io/lexicon/content/icons-glyphicons/}
```
jbalsas: All these links will soon disappear and I don't thin they'll be in the lexicon site. Should we simply point to FontAwesome and Glyphicons documentation directly?

packages/metal-quartz-icon/src/MetalQuartzIcon.soy
```diff
@@ -3,15 +3,56 @@
...
 	{if isNonnull($symbol) and isNonnull($spritemap)}
-		<svg aria-hidden="true" class="lexicon-icon lexicon-icon-{$symbol}{$elementClasses ? ' ' + $elementClasses : ''}">
-			<use xlink:href="{$spritemap}#{$symbol}" />
-		</svg>
+		{let $quartzIconElementClasses kind="text"}
+			{$elementClasses ? ' ' + $elementClasses : ''}
+		{/let}
+
+		{let $quartzIconVariant kind="text"}
+			{$spritemap == 'lexicon-fa' ? 'lexiconFa' : $spritemap}
```
jbalsas: When will the variant become `glyphicons`? I'm assuming it gets triggered if `spritemap=='glyphicons'`. If so, why don't do the same for `spritemap==fontawesome`?

packages/metal-quartz-icon/src/MetalQuartzIcon.soy
```diff
@@ -3,15 +3,56 @@
...
+{deltemplate QuartzIcon variant="'glyphicons'"}
+	<span aria-hidden="true" class="glyphicon {$symbol}{$quartzIconElementClasses}"></span>
+{/deltemplate}
+
+ /**
+ * @param? quartzIconElementClasses
+ * @param spritemap
+ * @param symbol
+ */
+{deltemplate QuartzIcon variant="'lexiconFa'"}
```
jbalsas: I think I'd rather call this simply `fontawesome`,  similar to `glyphicons`. What do you think?

jbalsas (changes requested):
Hey @pat270, thanks a lot for working on this! What do you think about it so far? 

One big question about this I have is the fontawesome and glyphicons support... that implies that Quartz will always be based on bootstrap and/or contain those. It's fine if that's what we want, just trying to clarify it.

packages/metal-quartz-icon/src/MetalQuartzIcon.soy
pat270: See above.

packages/metal-quartz-icon/demos/index.html
pat270: So far it hasn't been that hard to learn (it helps to have all the other metal modules to look at), it's just a lack of metal documentation that increases the learning curve a bit. I don't think it's possible to decouple Bootstrap from these modules since we aren't including any css. The only way it makes sense is if we remove Bootstrap from Quartz.

I can change this to Quartz Fa.

packages/metal-quartz-icon/src/MetalQuartzIcon.js
pat270: We can link the Glyphicon documentation directly, but we still use the old version of Font Awesome that we have customized from 6.2. We never migrated to Font Awesome 4.

packages/metal-quartz-icon/src/MetalQuartzIcon.soy
pat270: See above.

robframpton:
I can see a value for maintaining font-awesome since it's still widely used, even if we don't base off of Bootstrap in the future. Not sure about Glyphicons, @pat270 what do you think?

jbalsas:
I'm fine with keeping one or both, I just wanted to point out that if we do that, we'll need to keep that support around for versions, or be careful whenever we want to remove it, because it will be a major breaking change...

packages/metal-quartz-icon/src/MetalQuartzIcon.soy
```diff
@@ -1,17 +1,57 @@
...
-		<svg aria-hidden="true" class="lexicon-icon lexicon-icon-{$symbol}{$elementClasses ? ' ' + $elementClasses : ''}">
-			<use xlink:href="{$spritemap}#{$symbol}" />
-		</svg>
+		{let $quartzIconElementClasses kind="text"}
+			{$elementClasses ? ' ' + $elementClasses : ''}
+		{/let}
+
+		{let $quartzIconVariant kind="text"}
+			{$spritemap == 'quartz-fa' ? 'quartzFa' : $spritemap}
+		{/let}
```
jbalsas: Hey @pat270, can't we just assign `$quartzIconVariant` to `$spritemap`? Then, you can simply pass `quartzFa`, `glyphicons` or $spritemap.

Also, `quartzFa` might not be very descriptive...

Another option would be to have an additional `@param iconLibrary` that could be `undefined`, `quartz` or `glyphicons`, so we'd have a more explicit API. What do you think? @Robert-Frampton?

jbalsas: Hey @pat270, can't we just assign `$quartzIconVariant` to `$spritemap`? Then, you can simply pass `quartzFa`, `glyphicons` or $spritemap.

Also, `quartzFa` might not be very descriptive...

Another option would be to have an additional `@param iconLibrary` that could be `undefined`, `quartz` or `glyphicons`, so we'd have a more explicit API. What do you guys think, @pat270, @Robert-Frampton, @brunobasto?

jbalsas (changes requested):
Hey guys, this looks mostly good to me, I just have some concerns over the `spritemap` API.

What do you think about having an additional `iconLibrary`, `iconFont` so we could pick `quartz(FontAwesome)` or `glyphicons`?

By default we won't pass anything here, so we'll use the `spritemap` param to render the custom svg icons in quartz...

How does that sound? Do you have any other suggestion for this?

packages/metal-quartz-icon/src/MetalQuartzIcon.soy
pat270: @jbalsas, I'm fine with adding another parameter, then we can pass in any icon font library.

robframpton: Hey @jbalsas 

You actually can't reassign params in soy, and you can't reassign `let` values either.

+1 for `@param iconLibrary`.

I also don't think `fa` is too big of a problem since font awesome itself uses it all the time.

`<i class="fa fa-address-book" aria-hidden="true"></i>`

pat270: Hey @jbalsas @Robert-Frampton, what do you guys think about doing away with the extra param, glyphicons, and font awesome and just allow the user to pass in class names through elementClasses. This way we can use any icon library that uses a span tag for the markup.

jbalsas: Hey @pat270, would that mean that the developer will need to pass `symbol: glyphicon glyphicon-arrow` or similar? or should we do that via `elementClasses`?

When will we generste the span indtead of the svg sprite syntax? Whenever the developer didn't pass `spritemap`?

pat270: @jbalsas, yes they would pass in the classes via `elementClasses` and span will be generated when there is no `spritemap` or `symbol`.

packages/metal-quartz-icon/src/MetalQuartzIcon.js
```diff
@@ -20,25 +19,33 @@ class MetalQuartzIcon extends Component {
...
 MetalQuartzIcon.STATE = {
 	/**
-	 * Path to the spritemap file with the icons.
+	 * CSS classes to be applied to the element, use this to pass in classes for
+	 * third-party icon libraries e.g. `glyphicon glyphicon-plus`. If
+	 * elementClasses is defined and spritemap or symbol is undefined, it will
+	 * insert elementClasses into a span tag.
+	 * @type {string}
+	 */
+	elementClasses: {
```
robframpton: `elementClasses` is actually a default property on `Component`, so we don't need to add it to anything extending from it.

packages/metal-quartz-icon/src/MetalQuartzIcon.js
```diff
@@ -20,25 +19,33 @@ class MetalQuartzIcon extends Component {
...
+	/**
+	 * The path to the SVG spritemap file containing the icons.
 	 * @instance
 	 * @memberof MetalQuartzIcon
 	 * @type {string}
 	 */
 	spritemap: {
-		required: true,
-		validator: core.isString
+		validator: value => core.isString(value)
```
robframpton: It should work fine with just `validator: core.isString`, that's what we do in a lot of metal components. Did this result in different behavior?

pat270: I actually don't know what validator does. I'm just copying what I saw in other metal components. It works even if I don't declare it or if I declare it and pass in an object or array.

robframpton:
Hey @pat270 

This looks good to me, the only change I would make is manually adding the `elementClasses` to the markup. Even though Metal does it automatically, I think we should leave it for server side rendering.

robframpton (approved):

PR #8, opened 2017-03-15
WIP - Metal Quartz Button Component

fernandosouza:
Hey, guys. This is still working in progress, but I will like to put it on evaluation before continue. I think we already have a considerable amount of code for reviewing.

We can extract a base implementation and create a `metal-button` component.

cc. @brunobasto


packages/metal-quartz-button/src/MetalQuartzButtonAttributes.soy
```diff
@@ -0,0 +1,36 @@
...
+	{@param? size: string}
+	{@param? style: string}
+	{@param? type: string}
+	{@param? value: string}
+
+	{let $blockClass: $block ? ' btn-block' : '' /}
+	{let $sizeClass: $size ? ' btn' + $size : '' /}
+	{let $styleClass: $style ? ' btn-' + $style : ' btn-default' /}
+
+	class="btn{$blockClass}{$styleClass}{$sizeClass}"
```
fernandosouza: I would like to find a way to allow users to pass any attribute name to the component. Have you guys any idea?

packages/metal-quartz-button/test/Button.js
```diff
@@ -0,0 +1,68 @@
...
+			sample.dispose();
+		}
+	});
+
+	it('should generate the default markup', function() {
+		sample = new MetalQuartzButton();
+
+		assert.strictEqual(sample.element.outerHTML, DEFAULT_BUTTON);
+	});
+
```
brunobasto, line 28: @fernandosouza can you run "gulp format" to remove unnecessary spaces?

packages/metal-quartz-button/src/MetalQuartzButtonAttributes.soy
```diff
@@ -0,0 +1,36 @@
...
+
```
brunobasto: I think "gulp format" does not remove spaces form .soy files?

packages/metal-quartz-button/src/MetalQuartzButtonBase.js
```diff
@@ -0,0 +1,91 @@
...
+	 * @type {boolean}
+	 * @default false
+	 */
+	disabled: {
+		validator: core.isBoolean,
+		value: false
+	},
+
+	/**
+	 * The link href.
```
brunobasto: @fernandosouza I think we could elaborate a little bit more on the jsdoc here. 
@jbalsas, I know we are not matching 1 to 1 on features, but href (buttons that navigate when clicked) is something we wanna keep, right?

packages/metal-quartz-button/src/MetalQuartzButtonAttributes.soy
brunobasto: In fact the Incremental DOM compiler is complaining. Let's try to think of something.

packages/metal-quartz-button/test/Button.js
fernandosouza: I did yesterday. There is a source format commit.

packages/metal-quartz-button/src/MetalQuartzButtonAttributes.soy
fernandosouza: Source formating commit updated

packages/metal-quartz-button/src/MetalQuartzButtonBase.js
jbalsas: Yeah, I think so... this was our old implementation and it had those as well: https://github.com/liferay-lexicon/lexicon-button

This should be the relevant document now: https://liferay.invisionapp.com/d/main#/console/10324888/219580168/preview

packages/metal-quartz-button/src/MetalQuartzButton.soy
```diff
@@ -0,0 +1,47 @@
...
+	{@param? label: html}
+	{@param? name: string}
+	{@param? size: string}
+	{@param? style: string}
+	{@param? type: string}
+	{@param? value: string}
+	<a href="{$href or ''}" {call MetalQuartzButtonAttributes.attributes data="all" /}>
+		{$label or ''}
+	</a>
+{/deltemplate}
```
jbalsas: Hey @fernandosouza, I feel the use of `deltemplate` here makes everything way too verbose for my taste. Why don't just have an if with different `call` expressions? It is usually easier to reason with, and people don't need to learn about `deltemplates.

What do you think?

packages/metal-quartz-button/src/MetalQuartzButton.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+class MetalQuartzButton extends MetalQuartzButtonBase {
+}
+
+/**
+ * State definition.
+ * @static
+ * @type {!Object}
+ */
+MetalQuartzButton.STATE = {
```
jbalsas, line 22: We're probably missing something like `enabled/disabled`?

jbalsas: What about icons? We can reuse the already existing [`quartz-icon` component](https://github.com/metal/metal-quartz-components/tree/master/packages/metal-quartz-icon)

jbalsas (changes requested):
Hey @fernandosouza, I know this is still wip. I think the think that I'm not so sure about is the usage of `deltemplates`.

One of the keys of using a template language such as `Closure Templates` is that it so closely resembles `HTML`. This greatly reduces the entry barrier for non-developers and will drive more contributors to the project.

packages/metal-quartz-button/src/MetalQuartzButton.js
fernandosouza: Hey, @jbalsas. We have a disable attribute here. https://github.com/metal/metal-quartz-components/pull/8/files/d2e9cf1d793f67c445174f03087c8441cdb795dc#diff-68f8bcc3875d81fc06327e0bcdff3b00R25

But yeah, we are missing some features like `icons` for example. I've preferred to make an initial pull request to get your thoughts before continue.

packages/metal-quartz-button/src/MetalQuartzButton.soy
fernandosouza: Looking from the point of view of non-developers, yes, makes sense. An `if` statement is more readable than `deltemplates`

packages/metal-quartz-button/src/MetalQuartzButton.soy
```diff
@@ -0,0 +1,40 @@
...
+		</a>
+	{else}
+		<button {call MetalQuartzButtonAttributes.attributes data="all" /}>
+			{call .iconMarkup data="all" /}
+			{$label or ''}
+		</button>
+	{/if}
+{/template}
+
+{template .iconMarkup}
```
robframpton: I think for our template naming convention, we should avoid using the work "markup", since all templates are realistically going to be returning some sort of markup. I would call this just `{template .icon}`.

packages/metal-quartz-button/src/MetalQuartzButtonAttributes.soy
```diff
@@ -0,0 +1,36 @@
+{namespace MetalQuartzButtonAttributes}
+
+/**
+ * This renders the component's whole content.
```
robframpton: Maybe change to something like "Renders html attributes for MetalQuartzButton component"

robframpton (changes requested):

fernandosouza:
Branch updated @Robert-Frampton @brunobasto @jbalsas

robframpton (approved):

packages/metal-quartz-button/test/.jshintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "mocha": true,
+  "globals": {
+    "assert": true
```
brunobasto: can you add `__html__` to globals so that "gulp lint" doesn't fail?

fernandosouza: Sure!

fernandosouza: Done.

brunobasto (approved):

PR #10, opened 2017-03-21
Create Metal Quartz Collapse Component

pat270:
https://github.com/metal/metal-quartz-components/issues/3


packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
```diff
@@ -0,0 +1,273 @@
...
+
+	/**
+	 * Toggles the content's visibility.
+	 */
+	toggle() {
+		this.transitionEventHandler_.add(
+			dom.delegate(document, this.transitionEnd, '.' + this.transitionClass, this.handleTransitionEnd_.bind(this))
+		);
+
+		this.collapsed = this.collapsed ? false : true;
```
fernandosouza: Could we simplify this for `this.collapsed = !this.collapsed;`? What do you think?

brunobasto:

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
```diff
@@ -1,15 +1,209 @@
...
+	syncCollapsed() {
+		this.collapsed ? this.animateClose() : this.animateOpen();
+	}
+
+	/**
+	 * Syncs the component according to the value of the `headers` state,
+	 * attaching events to the new element and detaching from any previous one.
+	 */
+	syncHeaders() {
+		this.eventHandler_.removeAllListeners();
```
brunobasto, line 183: Is it possible to do event delegation instead of individual click and keydown events for each node?

brunobasto (changes requested):

fernandosouza:

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
```diff
@@ -0,0 +1,273 @@
...
+
+	/**
+	 * Animates open when `collapsed` is false
+	 */
+	animateOpen() {
+		this.content.transitionType = 1;
+
+		dom.removeClasses(this.content, this.closedClass);
+		dom.addClasses(this.content, this.transitionClass);
+		this.content.setAttribute('style', 'height: ' + this.content.firstElementChild.offsetHeight + 'px;');
```
fernandosouza: We have the same code here and in the `animateClose()` method. What do you think about creating a private method named, for example, `updateContentHeight_`?

What do you think about using Template literals here?

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
```diff
@@ -0,0 +1,273 @@
...
+
+		this.transitionEventHandler_.removeAllListeners();
+	}
+
+	/**
+	 * Adds CSS classes and properties to the `content` element when `collapsed`
+	 * is false
+	 */
+	open() {
+		dom.addClasses(this.content, this.closedClass);
```
fernandosouza: Shouldn't we remove the `closedClass` here?

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
```diff
@@ -0,0 +1,273 @@
...
+		this.content.setAttribute('style', 'height: ' + this.content.firstElementChild.offsetHeight + 'px;');
+
+		this.shimUnsupportedTransition_(this.content);
+	}
+
+	/**
+	 * Adds CSS classes and properties to the `content` element when `collapsed`
+	 * is true
+	 */
+	close() {
```
fernandosouza: Both public methods `open()` and `close()` don't update `this.collapsed`. Its value might not reflect the current component state.

fernandosouza (changes requested):

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
```diff
@@ -0,0 +1,273 @@
...
+				dom.on(this.headers, 'keydown', this.handleKeydown_.bind(this))
+			);
+		}
+	}
+
+	/**
+	 * Toggles the content's visibility.
+	 */
+	toggle() {
+		this.transitionEventHandler_.add(
```
fernandosouza, line 204: Is it necessary to add a new event listener whenever toggle() is called?

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
pat270: You can pass in a css selector, a dom element, or an array of selectors and dom elements. It looks like Metal's `dom.on` function delegates it on the document for you if it's a css selector. I'm not sure if there's a way to delegate dom elements with Metal.

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
pat270: I can change this.

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
pat270: I can extract the height thing to its own method.

I'm not sure how to integrate template literals here, this component doesn't have a template.

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
pat270: close(), open(), animateClose(), animateOpen() aren't meant to be used outside the plug in, I can denote them as private.

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
pat270: Bootstrap's open state requires the classes `collapsed in` I'm not able to add more than one class at a time using Metal's `dom.addClass`. I can create a work around.

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
pat270: This component can be instantiated multiple times on a page. If we add it on component creation, there will be multiple listeners on the same class. It will fire however many times it's been instantiated.

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
fernandosouza: You can use [ES6 template literal](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) to replace the string concatenation and make the code shorter.

```javascript
`height: ${this.content.firstElementChild.offsetHeight}px;`

File: packages/metal-quartz-collapse/src/MetalQuartzCollapse.js (hunk 4 above)
Comment (fernandosouza) [reply packages/metal-quartz-collapse/src/MetalQuartzCollapse.js]: Ok. It seemed a bit strange to me when I first look.

File: packages/metal-quartz-collapse/src/MetalQuartzCollapse.js (hunk 3 above)
Comment (pat270) [reply packages/metal-quartz-collapse/src/MetalQuartzCollapse.js]: I see, I can change this.

File: packages/metal-quartz-collapse/src/MetalQuartzCollapse.js (hunk 4 above)
Comment (pat270) [reply packages/metal-quartz-collapse/src/MetalQuartzCollapse.js]: Just an update: I can add and remove more than one class with `dom.addClasses`, `dom.removeClasses`. I was using those with `dom.hasClass` previously and was running into issues. I rolled back the work around.

Comment (fernandosouza) [reply packages/metal-quartz-collapse/src/MetalQuartzCollapse.js]: I guess the issue is because `dom.hasClass` accept only one class name.

**Comment by fernandosouza** [comment]

File: packages/metal-quartz-collapse/src/MetalQuartzCollapse.js (hunk 6 above)
Comment (fernandosouza) [reply packages/metal-quartz-collapse/src/MetalQuartzCollapse.js:204]: Could it be changed to listen to events from `this.content` instead of `this.transitionClass`?

```javascript
this.transitionEventHandler_.add(
    dom.delegate(document, this.transitionEnd, this.content, this.handleTransitionEnd_.bind(this))
);
```

By evaluating the DOM element itself instead of a class name, we can move this code to the class initialization method.

What do you think about it, Patrick?

fernandosouza:
Hey, @pat270.

I don't know if it is a really important change, but some methods aren't annotated with its access level, (`@public`, `@private` or `@protected`).

Is there any default option that we can assume when there is no annotation?

packages/metal-quartz-collapse/src/MetalQuartzCollapse.js
pat270: We don't want open_() and close_() to run on every transitionend. We will run into problems if there is another kind of transition added to the content. It will open or close the menu after that transition as well.

From a performance point of view, I don't know if having X transitionend listeners on X numbers of content is actually better than adding and removing one on toggle.

fernandosouza: Hey, @pat270. Sorry for the delay. I understood the reason.

fernandosouza (approved):

fernandosouza:
LGTM

robframpton:
Hey guys, I made some formatting changes and resent (#15).

PR #12, opened 2017-03-21
Consolidates button templates

robframpton:
Having the `attributes` template in it's own file was causing issues when trying to render this component in another component's soy file.

@fernandosouza was there a reason for separating them that I'm not aware of?


pat270 (approved):

jbalsas (approved):
This looks good to me too. I was goimg to ask the other day, but it slipped my mind. We need to make these as closer to html as possible to maximize the possible contributor pool and minimize the template learning curve :)

PR #13, opened 2017-03-22
Validate button state using metal-state validators

robframpton:
`core.isString` and similar methods will not throw an error or notify the developer in any way if an improper value is passed to the state. 

I recommend using the metal-state validators for all quartz components going forward as they print very helpful error messages to the console.


jbalsas (approved):
LGTM

fernandosouza:
LGTM

fernandosouza (approved):

brunobasto (approved):
LGTM

Issue #18, opened 2017-03-28
Create Metal Quartz Link Component


pat270:
Reusable component that renders an anchor tag with a set list of attributes, will provide a consistent api across components that use links.

PR #19, opened 2017-03-30
Create Metal Quartz Link Component

pat270:
Issue #18

### DIFF
--- a/packages/metal-quartz-link/src/MetalQuartzLink.js
+++ b/packages/metal-quartz-link/src/MetalQuartzLink.js
@@ -20,17 +20,6 @@ class MetalQuartzLink extends Component {
  */
 MetalQuartzLink.STATE = {
 	/**
-	 * Sets the data-onclick attribute on the anchor tag.
-	 * @instance
-	 * @memberof MetalQuartzLink
-	 * @type {?string}
-	 * default undefined
-	 */
-	dataOnclick: {
-		validator: validators.string
-	},
-
-	/**
 	 * Sets the download attribute on the anchor tag.
 	 * @instance
 	 * @memberof MetalQuartzLink
--- a/packages/metal-quartz-link/src/MetalQuartzLink.js
+++ b/packages/metal-quartz-link/src/MetalQuartzLink.js
@@ -3,7 +3,6 @@
 import Component from 'metal-component';
 import MetalQuartzIcon from 'metal-quartz-icon';
 import Soy from 'metal-soy';
-import { core, object } from 'metal';
 import { validators } from 'metal-state';
 
 import templates from './MetalQuartzLink.soy.js';
@@ -12,12 +11,6 @@ import templates from './MetalQuartzLink.soy.js';
  * Implementation of the Metal Quartz Link.
  */
 class MetalQuartzLink extends Component {
-	/**
-	 * @inheritDoc
-	 */
-	created() {
-		MetalQuartzLink.instances.push(this);
-	}
 }
 
 /**
@@ -27,13 +20,24 @@ class MetalQuartzLink extends Component {
  */
 MetalQuartzLink.STATE = {
 	/**
-	 * Sets the data-toggle attribute on the anchor tag.
+	 * Sets the data-onclick attribute on the anchor tag.
 	 * @instance
 	 * @memberof MetalQuartzLink
 	 * @type {?string}
 	 * default undefined
 	 */
-	dataToggle: {
+	dataOnclick: {
+		validator: validators.string
+	},
+
+	/**
+	 * Sets the download attribute on the anchor tag.
+	 * @instance
+	 * @memberof MetalQuartzLink
+	 * @type {?string}
+	 * default undefined
+	 */
+	download: {
 		validator: validators.string
 	},
 
@@ -91,13 +95,19 @@ MetalQuartzLink.STATE = {
 	 */
 	label: {
 		isHtml: true
+	},
+
+	/**
+	 * Sets the HTML attribute target on the anchor tag.
+	 * @instance
+	 * @memberof MetalQuartzLink
+	 * @type {?string}
+	 * @default undefined
+	 */
+	target: {
+		validator: validators.string
 	}
 };
-
-/**
- * An array of all MetalQuartzLink's that are created on the page.
- */
-MetalQuartzLink.instances = [];
 
 Soy.register(MetalQuartzLink, templates);
 
--- a/packages/metal-quartz-link/src/MetalQuartzLink.js
+++ b/packages/metal-quartz-link/src/MetalQuartzLink.js
@@ -1,8 +1,10 @@
 'use strict';
 
 import Component from 'metal-component';
-import core from 'metal';
+import MetalQuartzIcon from 'metal-quartz-icon';
 import Soy from 'metal-soy';
+import { core, object } from 'metal';
+import { validators } from 'metal-state';
 
 import templates from './MetalQuartzLink.soy.js';
 
@@ -10,6 +12,12 @@ import templates from './MetalQuartzLink.soy.js';
  * Implementation of the Metal Quartz Link.
  */
 class MetalQuartzLink extends Component {
+	/**
+	 * @inheritDoc
+	 */
+	created() {
+		MetalQuartzLink.instances.push(this);
+	}
 }
 
 /**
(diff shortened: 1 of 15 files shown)


packages/metal-quartz-link/src/MetalQuartzLink.soy
```diff
@@ -0,0 +1,56 @@
...
+	{if $html}
+		{$html}
+	{else}
+		{let $attrs kind="attributes"}
+			{if $elementClasses}
+				class="{$elementClasses}"
+			{/if}
+
+			{if $dataToggle}
+				data-toggle="{$dataToggle}"
```
robframpton: I'm guessing this data attribute integrates with other lexicon components?

packages/metal-quartz-link/src/MetalQuartzLink.soy
```diff
@@ -0,0 +1,56 @@
...
+					{param elementClasses: $icon.elementClasses /}
+					{param spritemap: $icon.spritemap /}
+					{param symbol: $icon.symbol /}
+				{/call}
+			{/if}
+
+			{$label}
+
+			{if $icon and not $icon.alignment or $icon and $icon.alignment == 'right'}
+				{call MetalQuartzIcon.render data="all"}
```
robframpton: Let's pull this into another template in this file so we don't have to write it twice.

```
{template .icon}
  {call MetalQuartzIcon.render data="all"}
  ...
{/template}
```

packages/metal-quartz-link/src/MetalQuartzLink.js
```diff
@@ -0,0 +1,104 @@
+'use strict';
+
+import Component from 'metal-component';
+import MetalQuartzIcon from 'metal-quartz-icon';
+import Soy from 'metal-soy';
+import { core, object } from 'metal';
```
robframpton: Looks like these imports aren't being used anywhere.

packages/metal-quartz-link/src/MetalQuartzLink.soy
```diff
@@ -0,0 +1,56 @@
...
+	{@param? dataToggle: string}
+	{@param? elementClasses: string}
+	{@param? href: string}
+	{@param? html: html}
+	{@param? icon: ?}
+	{@param? label: html}
+	{@param? id: string}
+
+	{if $html}
+		{$html}
```
robframpton: If someone is passing in their own `html`, it seems like they might as well not use the component. Do you think we need this?

packages/metal-quartz-link/src/MetalQuartzLink.js
```diff
@@ -0,0 +1,104 @@
...
+	},
+
+	/**
+	 * Id to be applied to the anchor tag.
+	 * @instance
+	 * @memberof MetalQuartzLink
+	 * @type {?string}
+	 * default undefined
+	 */
+	id: {
```
robframpton, line 74: I'd have to check, but I think like `elementClasses`, this _might_ already be an attribute on the base `Component` class.

packages/metal-quartz-link/src/MetalQuartzLink.js
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Implementation of the Metal Quartz Link.
+ */
+class MetalQuartzLink extends Component {
+	/**
+	 * @inheritDoc
+	 */
+	created() {
+		MetalQuartzLink.instances.push(this);
```
robframpton: I'm curious what the use case of this would be? If it's something we need, we'll probably need to also remove it from the array in the `disposed` lifecycle method.

robframpton (changes requested):
Hey @pat270 , thanks for the pr. I've commented on a few small changes/things I wasn't sure about.

I was also wondering what we thought about adding a `target` and maybe `download` attribute, since those might be common enough to implement in this component.

packages/metal-quartz-link/src/MetalQuartzLink.soy
pat270: I think we need to provide this until we can find a way to provide custom attributes in soy. We will need to reimplement link in other components that might use attributes we haven't provided in this component.

packages/metal-quartz-link/src/MetalQuartzLink.soy
pat270: I put the wrong attribute in. It's supposed to be dataOnclick.

packages/metal-quartz-link/src/MetalQuartzLink.soy
pat270: I can change this.

packages/metal-quartz-link/src/MetalQuartzLink.js
pat270: I can change this.

packages/metal-quartz-link/src/MetalQuartzLink.js
pat270: This probably shouldn't go in here. I don't know if Metal has something that keeps track of components that have been rendered on the page. It's really difficult to access the state of a component. It's something I added to keep track of the components being rendered on the page so developers can have some kind of access.

packages/metal-quartz-link/src/MetalQuartzLink.js
pat270: I didn't see id in Component.

brunobasto: Hey guys, I though it had it too, but couldn't see it either.

packages/metal-quartz-link/src/MetalQuartzLink.js
brunobasto: Hey @pat270, from what I remember, from within a component you should be able to navigate in the tree of nested components rendered by the parent component by doing something like `this.components['icon']`

brunobasto: Although I just tried to access the icon component from within the button component and `this.components` is returning an empty map..

jbalsas: It is, one just needs to add the `ref` attribute to the components that should be exposed via the `this.components` API.

See https://github.com/metal/metal.js/blob/master/packages/metal-incremental-dom/src/render/render.js#L159-L165

pat270: Hey Guys, What I want to accomplish here is to expose the component to the outside world so that, lets say, someone using jQuery has access to the state without having to search through the window object. It would be nice to have some kind of method that returns a metal component on the page based on a selector, dom element, or some kind of attribute.

jbalsas: Hey @pat270, I'm not 100% sure about how good would that be. Depending on how the component re-renders, reference to subcomponents might get lost, outdated or cause inconsistent internal states.

If we do consider this to be implemented at all, I think it should be done in a general way (from within metal.js) with an API based on the `components.ref` information we have, or it can be left to the library user (Liferay Portal), but it should not be a per-component thing.

jbalsas: Could you write a use case? How do you think this could be used?

pat270: Lets say you have a component MetalQuartzList with some state called `order` that can be `alphabetical, date, or status`. I add some buttons that lets you change the `order` but I want to use jQuery instead of Metal. I want to be able to:

$('.my-button-alphabetical').on('click', function() {
    dom.getComponent('#myListId').order = 'alphabetical';
});

$('.my-button-date').on('click', function() {
    dom.getComponent('#myListId').order = 'date';
});

This is also useful for two unrelated Metal Components to be able to talk to each other.

fernandosouza (approved):

fernandosouza:
Once we have `metal-quartz-link` I think we can remove [this validation](https://github.com/metal/metal-quartz-components/blob/master/packages/metal-quartz-button/src/MetalQuartzButton.soy#L19) on `metal-quartz-button`.

cc. @pat270 @jbalsas @brunobasto @Robert-Frampton

packages/metal-quartz-link/src/MetalQuartzLink.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+MetalQuartzLink.STATE = {
+	/**
+	 * Sets the data-onclick attribute on the anchor tag.
+	 * @instance
+	 * @memberof MetalQuartzLink
+	 * @type {?string}
+	 * default undefined
+	 */
+	dataOnclick: {
```
robframpton: @pat270 and I already discussed it, but we can remove this attribute since metal has a way of adding DOM event handlers to components via the `events` attribute.

```javascript
new MetalQuartsLink({
  events: {
    click: function(event) {
    }
  }
});
```

robframpton (changes requested):

robframpton:
Hey @pat270 

I went and merged this with a few small changes. I would like to get rid of the `html` attribute as soon as possible so that we don't end up relying on it, so let's look into other potential solutions to the data attribute issue.

pat270:
@Robert-Frampton I don't think there's a solution to the data attribute issue. We can just get rid of it and see what happens.

robframpton:
Ok, I went ahead and sent a pr for it.

https://github.com/metal/metal-quartz-components/pull/26

Issue #21, opened 2017-03-31
Create Metal Clay Checkbox Component


pat270:

anitaderiane:
https://issues.liferay.com/browse/IFI-55

fernandosouza:
I am going to working on this one, @pat270.

fernandosouza:
Sorry @pat270. I've just seen you already sent a pull request.

carloslancha:
Since we now have https://github.com/metal/metal-clay-components/tree/master/packages/clay-checkbox can we close this issue?

PR #22, opened 2017-04-01
#21 Create Metal Quartz Checkbox Component

pat270:


packages/metal-quartz-checkbox/src/MetalQuartzCheckbox.js
```diff
@@ -19,6 +19,53 @@ class MetalQuartzCheckbox extends Component {
  * @type {!Object}
  */
 MetalQuartzCheckbox.STATE = {
+	checkboxInline: {
```
brunobasto: Hey @pat270,

From the name, it's not clear to me what this attribute does. So this is a list of objects with props for each checkbox?

pat270: Yes that is what it is, should look like when initializing from js:
```
checkboxInline: [{
	checked: true,
	label: 'Inline 1'
}, {
	label: 'Inline 2'
}, {
	checked: true,
	disabled: true,
	label: 'Inline 3'
}, {
	label: 'Inline 4'
}, {
	label: 'Inline 5'
}, {
	label: 'Inline 6'
}]
```
I was following what was being done at https://github.com/liferay-lexicon/lexicon-dropdown/blob/master/demos/index.html

robframpton:
Hey @pat270 , what do you think about creating a base component called `MetalQuartzInputBase` that we can use to house the common attributes that most input types will use? 

In particular I'm thinking `name`, `value`, `disabled`, `label`, etc. That way we don't have to redefine those attrs for every input component we create.

robframpton:
Due to f8abd7353e707302accfe8f6e644f3705ae98d61 this pr needs to be rebased.

fernandosouza:

jbalsas:
I'm closing this for now. @carloslancha can pick it up as starting point whenever we get back to this again!

PR #31, opened 2017-05-16
Metal Clay Button: Source formating

fernandosouza:


fernandosouza:
Do you, guys, think is it fine for merging?

packages/metal-clay-button/src/MetalClayButton.js
```diff
@@ -61,11 +61,10 @@ MetalClayButton.STATE = {
 	 * @instance
 	 * @memberof MetalClayButton
 	 * @type {?string}
-	 * @default 'default'
+	 * @default undefined
 	 */
 	style: {
 		validator: validators.string,
-		value: 'default'
```
brunobasto: Not sure this is an issue but looks odd to have the type declared as string but it can return undefined. I'd say we either update the  jsdoc or make the default value an empty string

packages/metal-clay-button/src/MetalClayButton.soy
```diff
@@ -17,27 +17,56 @@
...
+{template .content}
 	{@param? icon: ?}
+	{@param? label: html}
 
-	{if $icon}
-		{call MetalClayIcon.render}
-			{param elementClasses: $icon.elementClasses /}
-			{param spritemap: $icon.spritemap /}
-			{param symbol: $icon.symbol /}
+	{let $iconAlignment: $icon and $icon.alignment ? $icon.alignment : 'left' /}
```
brunobasto, line 54: Can we follow the default value guideline: https://github.com/liferay/com-liferay-osb-faro-build-private/wiki/Metal-Soy-Guidelines#default-values ?

packages/metal-clay-button/src/MetalClayButton.soy
```diff
@@ -17,27 +17,56 @@
...
+			{param icon: $icon /}
+		{/call}
+	{/if}
+{/template}
+
+/**
+ * Renders icon element
+ */
+{template .icon}
+	{@param icon: ?}
```
brunobasto: Shouldn't icon be of type Record?

brunobasto (changes requested):

packages/metal-clay-button/src/MetalClayButton.soy
fernandosouza: I'm not sure if it will work here once we have a logical operation for checking `$icon` and `$icon.alignment`.

packages/metal-clay-button/src/MetalClayButton.soy
fernandosouza: Yeah. I'm gonna change.

packages/metal-clay-button/src/MetalClayButton.js
fernandosouza: Probably will be nice to use the Config object mentioned here: https://github.com/liferay/com-liferay-osb-faro-build-private/wiki/Metal-Soy-Guidelines#declaring-params

packages/metal-clay-button/src/MetalClayButton.soy
brunobasto: I see. Then it should be fine.

packages/metal-clay-button/src/MetalClayButton.js
fernandosouza: I thought more about it and just update the JSDoc to `* @type {?string|undefined}` works. When we define an attribute value here that means that it is its default value, that is not the case here, there is no default value anymore, then makes sense return `undefined` here.

Does it make sense, @brunobasto?

brunobasto: I think so.

packages/metal-clay-button/src/MetalClayButton.soy
fernandosouza: @brunobasto, Michael alerts us for the use of `records`. The guideline says: 

> For records, there is a record syntax that would be useful, however it currently will lead to issues with your javascript component, so for the moment just use ?, which will still allow member access using the . operator. See this issue for the status for record type syntax for metal-soy.

brunobasto: Got it. Haven't seen that before.

jbalsas: This has been fixed in https://github.com/metal/metal-tools-soy/pull/11, so it should be available as soon as we release a new metal-cli version...

jbalsas (changes requested):
Hey @fernandosouza, while we're at it, we should use the `State.Config` validators instead of the simple `core` ones as a rule for these components as well.

fernandosouza:
Hey @brunobasto and @jbalsas. Branch updated.

pat270 (approved):

brunobasto (approved):
LGTM

fernandosouza:
Hey, @jbalsas. Does it look better now? :)

PR #36, opened 2017-06-08
MetalClayIcon: Add aria-label attribute

julien:


jbalsas:
Hey @julien, I'm going to close this for now since we've pushed significant changes in the components and tests, but please, do resend it if you think it is still relevant.

One thing I noticed is that none of the fixtures were actually showing an `aria-label` with content, so we either are missing some test case or it is not doing what I'd expect :)

julien:
@jbalsas ok thanks. I'll take a look at that with the changes and make sure to test it correctly.

jbalsas:
@julien you should be having 🍹 on the 🏖 !!

PR #37, opened 2017-06-08
MetalClaySample: Add role attribute for accessibility

julien:


jbalsas:
Hey @julien, I just removed this `sample`, so I guess we're no longer merging this, sorry! :/

PR #41, opened 2017-07-04
Removes `metal-` prefix from `metal-clay-components`

jbalsas:
Hey @Robert-Frampton, @brunobasto, I think we discussed this some time ago... so I just went ahead and removed the `metal-` prefix from the clay components. 

It should make them a bit easier to use and removes a lot of boilerplate.

What do you think?


robframpton (approved):
LGTM

PR #43, opened 2017-07-06
Fix Lerna install command

fernandosouza:


robframpton:
Hey @fernandosouza 

I think we can remove the version altogether now that they are on a normal release.

```
[sudo] npm install -g lerna
```

jbalsas:
Can you push the change, @fernandosouza?

fernandosouza:
Done. :)

PR #44, opened 2017-07-25
Clay Icon

carloslancha:


jbalsas:
Hey @Robert-Frampton, @pat270, @brunobasto, @marcoscv-work, @fernandosouza, we're starting to work with UX on this. We'll be referencing the LEXI tickets so we can trace it back to the UX definitions.

As a first step, we've agreed with UX that components should allow as little as possible, so we're going to trim them down and rebuild from there. 

What do you think?

PS: @carloslancha, we probably want to look into demos to update that as well and show only the existing features.

jbalsas:
LGTM, merging so we can move on, but feel free to add any comments, guys! :)

PR #48, opened 2017-07-27
LEXI-154 Clay Button

carloslancha:


carloslancha:
@marcoscv-work we need to add the class btn-borderless to clay. Thx!

PR #50, opened 2017-07-27
LEXI-174 ClayIcon

carloslancha:


packages/clay-icon/src/ClayIcon.js
```diff
@@ -18,6 +18,15 @@ class ClayIcon extends Component {}
  */
 ClayIcon.STATE = {
 	/**
+	 * Id to be applied to the element.
+	 * @instance
+	 * @memberof ClayButton
+	 * @type {?string|undefined}
+	 * default undefined
```
jbalsas: Should this be `@default`?

jbalsas (changes requested):

jbalsas (approved):

PR #54, opened 2017-08-22
Create Clay Select component

carloslancha:


packages/clay-select/src/ClaySelect.soy
```diff
@@ -0,0 +1,73 @@
...
+			{call .select}
+				{param disabled: $disabled /}
+				{param multiple: $multiple /}
+				{param name: $name /}
+				{param options: $options /}
+			{/call}
+		{/let}
+
+		{if $inline}
+			<div class="form-inline">
```
robframpton: Hey @carloslancha 

I believe the `form-inline` class is intended for a wrapper element like the `<form>` tag. I think we could possibly use the `form-inline-group` class on the root element instead.

@pat270 is working on updating all the markup documentation on the [claycss.com](https://github.com/liferay/claycss.com) repo, once he does we'll make it available internally so stuff like this is easier to find.

packages/clay-select/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+  ```
+  npm run build
+  ```
+
+4. Open the demo at demos/index.html on your browser.
+
+## Contribute
+
+We'd love to get contributions from you! Please, check our [Contributing Guidelines](CONTRIBUTING.md) to see how you can help us improve.
```
blzaugg, line 26: I'd like to see the API documented in `README.md`. Users should not need to parse the source (`STATE`, `demos/index.html`) to see how to use this.

jbalsas: Hey @blzaugg, API's will show up in the claycss.com site. Does it make sense to have them in here as well? If so, we need to find a way to auto-generate them so it is easier to maintain.

blzaugg: Good point. Let's link to the Clay site API when it's live.

packages/clay-select/src/ClaySelect.soy
carloslancha: Nice, as soon as claycss.com is updated I'll make the changes to fit them. Thx!

carloslancha:
@Robert-Frampton looking on lexicon patterns for select (which are the same that for inputs and textareas) there won't be inline option. The only possible thing could be to create a grid-like input groups, done by using `form-inline-group` instead `form-group`. I've updated necessary changes for this. Thx!

jbalsas:
@Robert-Frampton, are we good to merge this one?

PR #55, opened 2017-08-23
Create Clay Badge component

matuzalemsteles:


packages/clay-badge/src/ClayBadge.js
```diff
@@ -0,0 +1,128 @@
...
+	 * onClick in Label of the Badge or in all Badge.
+	 * @instance
+	 * @memberof ClayBadge
+	 * @type {?func|undefined}
+	 * default undefined
+	 */
+	onClickLink: Config.func(),
+
+	/**
+	 * Option to enable href on every badge.
```
brunobasto: Can you rephrase that? Maybe:

> Option to make the badge outmost container be an anchor (`<a href="">`).

Something like that.

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,122 @@
...
+		{if $id}
+			id="{$id}"
+		{/if}
+
+		{if $href}
+			href="{$href}"
+		{/if}
+	{/let}
+
+	{if $href and $linkWrapper}
```
brunobasto: From these checks, looks like I cannot have the whole badge wrapped around an anchor while at the same time have an icon displayed. Maybe it's an option we want people to be able to pick. What do you think?

brunobasto (changes requested):

packages/clay-badge/src/ClayBadge.js
matuzalemsteles: Yeah, I'll adjust.

robframpton (changes requested):
Hey @matuzalemsteles 

I think there are a few things we can simplify with this component.

Let's remove the `linkWrapper` attribute and just make the badge itself the outermost element, and make it either an anchor or a span based on the `href` attribute.

We can also remove the `onClick` attributes, metal allows us to attach event listeners to the root element of a component without having specified attributes.

For example:

```javascript
new metal.ClayBadge({
	label: 'Badge',
	events: {
		click: function(event) {
			console.log(event);
		}
	}
});
```

Let me know if you have any questions, thanks!

packages/clay-badge/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+  ```
+  npm run build
+  ```
+
+4. Open the demo at demos/index.html on your browser.
+
+## Contribute
+
+We'd love to get contributions from you! Please, check our [Contributing Guidelines](CONTRIBUTING.md) to see how you can help us improve.
```
blzaugg, line 26: I'd like to see the API documented in `README.md`. Users should not need to parse the source (`ClayBadge.STATE`, `demos/index.html`) to see how to use this.

matuzalemsteles: Hey @blzaugg, I'll work on it.

matuzalemsteles:
Hey @Robert-Frampton,

Events of the metal facilitates quite really, but of the big concern was the question of cases of use of the Liferay. Uses events and links.

The events API really makes it cleaner, so I think we can just use that. The only note is that, if some one wants to attach two click events, one for the icon and another for the badge itself, they would have two options:

1. Use just one click for the badge container and check if the event target is the container or the icon to decide what action to take.
2. Do something like:

```javascript
new metal.ClayBadge(
  {
    events: {
      click: handleClickBadge
    },
    label: 'Badge',
    icon: {
      ...
      events: {
        click: handleClickIcon
      }
    },
  },
  ...
);
```

I tried option 2, but it doesn’t seem like the Icon subcomponent will receive the events when they get instantiated by the soy `{call}`

> Let's remove the `linkWrapper` attribute and just make the badge itself the outermost element, and make it either an anchor or a span based on the href attribute.

I added the `linkWrapper` option (I agree that the name needs improvement) cause I wanted to follow all the listed alternatives on the spec. [If you look here](http://pat270.github.io/lexicon/content/badges-and-labels/): we can see that there’s an option called `Badges as Anchor Tag` and another one called `Badge with Links`. So, I was just trying to implement whatever the spec says. But I understand that it could be too many options for a simple component like that. So, do you think we should stick with having just the first option and forget about the second use case?

robframpton:
Hey @pat270 

Do the invision designs have the link inside the badge pattern? If not we can remove it from this component.

For reference:

screen shot 2017-08-23 at 5 01 56 pm

pat270:
The Invision designs don't have them for badges, they are in labels though.

packages/clay-badge/README.md
jbalsas: Hey @blzaugg, API's will show up in the claycss.com site. Does it make sense to have them in here as well? If so, we need to find a way to auto-generate them so it is easier to maintain.

brunobasto:
So, should we just go with the link as a wrapper?

packages/clay-badge/README.md
blzaugg: Good point. Let's link to the Clay site API when it's live.

robframpton:
Hey @brunobasto @matuzalemsteles , given that the design patterns don't have that use case, let's go ahead and remove it.

If it ends up being something we need, then we can always add that functionality later, but we wouldn't be able to remove it if we added it now. We want to keep this components as high level implementations without too many options.

In regards to the events, you can actually accomplish this behavior already.

```javascript
new clay.Badge({
    label: 'Badge',
    events: {
        click: {
            selector: '.some-class',
            fn: function() {
                console.log('Hello, World!');
            }
        }
    }
});
```

You can do the same thing through a soy param as well

```
{param events: ['click': ['selector': '.some-class', 'fn': 'close']] /}
```

matuzalemsteles:
@Robert-Frampton, 

> If it ends up being something we need, then we can always add that functionality later, but we wouldn't be able to remove it if we added it now. We want to keep this components as high level implementations without too many options.

I agree with you.

> In regards to the events, you can actually accomplish this behavior already.
> ```javascript
> new clay.Badge({
>     label: 'Badge',
>     events: {
>         click: {
>             selector: '.some-class',
>             fn: function() {
>                 console.log('Hello, World!');
>             }
>         }
>     }
> });
> ```
> You can do the same thing through a soy param as well
> 
> `{param events: ['click': ['selector': '.some-class', 'fn': 'close']] /}`

Perfect, I'll send the changes.

packages/clay-badge/src/ClayBadge.js
```diff
@@ -0,0 +1,101 @@
...
+	href: Config.string(),
+
+	/**
+	 * Badge style color.
+	 * @instance
+	 * @memberof ClayBadge
+	 * @type {?string|undefined}
+	 * default undefined
+	 */
+	style: Config.oneOf([
```
carloslancha, line 86: Should be alphabetically ordered

packages/clay-badge/src/ClayBadge.js
```diff
@@ -0,0 +1,101 @@
...
+	size: Config.oneOf(['sm', 'lg']),
+
+	/**
+	 * Shape of the badge.
+	 * @instance
+	 * @memberof ClayBadge
+	 * @type {?string|undefined}
+	 * default undefined
+	 */
+	shape: Config.oneOf(['pill']),
```
carloslancha, line 68: As I can see on Badges Clay docs there're no shapes for Badges, only that they can be pills or not. Maybe would be better to not use `shape` parameter for this but a boolean `pill` one.

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,110 @@
...
+{template .render}
+	{@param? icon: [
+		alignment: string,
+		monospaced: bool,
+		spritemap: string,
+		symbol: string
+	]}
+	{@param? id: string}
+	{@param? label: html|string}
+	{@param? href: string}
```
carloslancha, line 15: Should be alphabetically ordered

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,110 @@
...
+			id="{$id}"
+		{/if}
+
+		{if $href}
+			href="{$href}"
+		{/if}
+	{/let}
+
+	{if $href}
+		<a {$attributes}>
```
carloslancha, line 43: Why don't use ClayLink here?

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,110 @@
...
+		alignment: string,
+		monospaced: bool,
+		spritemap: string,
+		symbol: string
+	]}
+	{@param? label: html|string}
+
+	{let $iconAlignment: $icon and $icon.alignment ? $icon.alignment : 'right' /}
+
+	{if $icon and $iconAlignment == 'left'}
```
carloslancha, line 74: As I can see in Lexicon there's no option for icons in badges to be on the left side of the label. In fact, the only accepted icon is the closing one so maybe we can here remove the icon stuff from the API and use a boolean like `closeable` or something like that and, according to this, render or not the close ClayIcon on the right side of the label. What do you think?

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,110 @@
...
+			{call .icon}
+				{param icon: $icon /}
+			{/call}
+		</button>
+	{/if}
+
+	{$label ? $label : ''}
+
+	{if $icon and $iconAlignment == 'right'}
+		<button aria-label="close" class="close" type="button">
```
carloslancha, line 85: Why don't use ClayButton here?

packages/clay-badge/src/ClayBadge.soy
brunobasto: Hey @carloslancha, I'm not sure about that. Sounds limiting, but as Rob pointed out, we can always add more stuff later if we need.

jbalsas: As @carloslancha is hinting, we've been asked several times by UX to keep the components and their variations as limited and restricted as possible. If it isn't designed or approved by the Lexicon team, there's no point in providing the flexibility in the components.

I agree that it's better to start small, with way simpler APIs and then grow from there.

brunobasto: Got it. Makes sense.

packages/clay-badge/src/ClayBadge.soy
matuzalemsteles: Hey @carloslancha,

It would be interesting, but ClayLink is not reusable. I would not be able to customize the `class`.

packages/clay-badge/src/ClayBadge.soy
matuzalemsteles: Hey @carloslancha,

ClayButton is not reusable. I would not be able to customize the `class`.

packages/clay-badge/src/ClayBadge.soy
matuzalemsteles: Hey @carloslancha @Robert-Frampton,

With the use of `closeable` I would have to add the svg url in the code, that would be a bad practice in my vision, you would have some idea?

```html
{if $closeable}
    <button aria-label="close" class="close" type="button">
        {call ClayIcon.render}
            {param monospaced: false /}
            {param spritemap: '../node_modules/clay/build/images/icons/icons.svg' /}
            {param symbol: 'times' /}
        {/call}
    </button>
{/if}
```

carloslancha: Hey @matuzalemsteles I'd say we need to pass the spritemap to the component as usual

packages/clay-badge/src/ClayBadge.soy
carloslancha: You're right, we probably need to add `elementClasses` param to ClayButton.

packages/clay-badge/src/ClayBadge.soy
carloslancha: You're right, we probably need to add `elementClasses` param to ClayLink.

packages/clay-badge/src/ClayBadge.soy
robframpton: Yeah, we should just add `elementClasses` to it.

robframpton: @matuzalemsteles Note that all metal components already have an `elementClasses` attribute inherited from `Component`, but we would still need to add it as a soy param for server side rendering.

carloslancha: https://github.com/metal/metal-clay-components/pull/56

packages/clay-badge/src/ClayBadge.soy
matuzalemsteles: @carloslancha Is it not better to let `icon`. Pass as before? Since we will use `ClayLink` and `ClayButton`.

carloslancha: @matuzalemsteles Since we don't want to provide the feature of use any icon but close one, pass an icon object as paran would be confusing IMHO.

packages/clay-badge/src/ClayBadge.soy
robframpton: #56 has been merged.

PR #56, opened 2017-08-25
Add elementClasses to ClayLink and ClayButton component

carloslancha:


carloslancha:
cc @matuzalemsteles this pr is to add the elementClasses to ClayLink and ClayButton. Once merged you should be able to use this components on the Badges one ;)

packages/clay-button/src/__tests__/__snapshots__/ClayButton.js.snap
```diff
@@ -4,6 +4,8 @@ exports[`ClayButton should generate the default markup 1`] = `<button class="btn
 
 exports[`ClayButton should render a block button 1`] = `<button class="btn btn-block btn-primary"></button>`;
 
+exports[`ClayButton should render a button with custom classes 1`] = `<button class="btn my-custom-class btn-primary"></button>`;
```
matuzalemsteles, line 7: Hey @carloslancha wouldn't it be better to just leave `my-custom-class`? If I use this along with the badge, it may break because of the `btn btn-primary`, do not you think?

carloslancha:
@matuzalemsteles It shouldn't, the correct use un badges should be with link style

packages/clay-button/src/ClayButton.js
```diff
@@ -39,13 +39,22 @@ ClayButton.STATE = {
...
 	disabled: Config.bool().value(false),
 
 	/**
+	 * CSS classes to be applied to the element.
+	 * @instance
+	 * @memberof ClayButton
+	 * @type {?string|undefined}
+	 * @default false
+	 */
+	elementClasses: Config.string(),
```
robframpton, line 55: Just as a note, technically these components will already inherit the `elementClasses` prop from https://github.com/metal/metal.js/blob/master/packages/metal-component/src/Component.js#L664-L672, but I actually like the visibility of having it here. API docs generation in the future will also be easier this way.

robframpton (approved):
LGTM.

packages/clay-button/src/ClayButton.js
carloslancha: Hey @robert-frampton yep, we did It in this way in FARO because of that 😃

PR #57, opened 2017-08-28
Fix package.json

matuzalemsteles:


brunobasto (approved):
Just to give context, this was breaking webpack build when depending on one of these modules.

LGTM.

robframpton:
Hey guys,

I believe this is what the [mainFields](https://webpack.js.org/configuration/resolve/#resolve-mainfields) setting is for in webpack.

I believe something like this would work.

```javascript
mainFields: ['jsnext:main', 'main']
```

jbalsas:
In node-world, `main` would usually signal a node (cjs) module, which is what would live in the `lib` folder. 

We have pure ES6 in the `src` folder, so that entry point is signaled in `jsnext:main`, although I think there's some discussion to rename it to `module` or `esnext`...

In any case, @Robert-Frampton's solution sounds like the right approach

matuzalemsteles:
Hey @Robert-Frampton @jbalsas ,

We can do that, but we'll have to make a change in the **metal-lerna-jest-resolver**, so he can solve the clay components in the jest and get him to read the `mainFields` of the webpack. We will have to use `esnext: main` because of Webpack 2.

It has this solution to solve the components of clay with `esnext: main`. Strange putting `esnext: main` right in the code, I suggest we get the` mainFields` from `webpack.config.js`. What do you think?
```javascript
const relativeDir = path.relative(path.resolve(process.cwd(), 'packages'), data.basedir);
    
if (file === 'metal-soy-bundle' || file === 'metal-incremental-dom' || file === 'incremental-dom') {
	data.basedir = path.resolve(process.cwd(), 'packages', relativeDir.split(path.sep)[0], 'node_modules', 'metal-soy');
}

if (file.indexOf('clay-') === 0) {
	data.basedir = path.resolve(process.cwd(), 'packages', relativeDir.split(path.sep)[0]);

	data.packageFilter = function (package) {
		return { main: package['esnext:main'] };
	};
}

return resolve.sync(file, data);
```

robframpton:
Hey @matuzalemsteles,

Could we do something more general in the jest resolver rather than check for the `clay-` prefix? That way it can work for our other projects as well. Maybe we could check to see what fields exist in the `package.json` and resolve accordingly?

matuzalemsteles:
@Robert-Frampton ,

I had to use the prefix `clay-`, because when we have equal dependencies between the packages. Example:
* ClayBadge
  * ClayIcon
  * ClayLink
    * ClayIcon

Jest accuses:
```
Namespace "ClayIcon.incrementaldom" already declared.
```

I see it better to get the `mainFields` that pass through` webpack.config.js`, so we can build an integration between them. I think catching `package.json` would have to know who the files are to start, if it does not follow the pattern. What do you think?

robframpton:
@matuzalemsteles 

Maybe for now, can we put this resolver into this repo somewhere so at least the tests will run and people can keep working? Then we can think of a more general fix that will work for other projects that have a similar structure.

matuzalemsteles:
@Robert-Frampton

I agree, we can leave this as a temporary solution in another repo not to mess up the **metal-lerna-jest-resolver**. what do you think ?

robframpton:
Sounds good, we can call it `clay-lerna-jest-resolver`.

matuzalemsteles:
Great, I will do this and send another pr with these updates.

jbalsas:
Hey guys, I'm looking to test a general approach to fix this, but I don't know where to find a code example for it. The idea is to simply cache the modules data so we always reuse the same:

```javscript
let cache = {};

module.exports = function (file, data) {
    // The shouldFilter check can be implemented by passing globs inside package.json
    // like `resolverDedupeGlobs: ['clay-', 'metal-']
    if (shouldFilter(file)) {
        const cachedFileData = cache[file];

        if (!cachedFileData) {
            // This might be slightly different for the `lerna` case
            data.basedir = cwd;
            cache[file] = data;
        }

        data = cache[file];
    }

    return resolve.sync(file, data);
}
```

@brunobasto, @Robert-Frampton, @matuzalemsteles, thoughts?

robframpton:
Hey @jbalsas, sounds good to me.

PR #59, opened 2017-08-30
Adding clay-lerna-jest-resolver

matuzalemsteles:
Adding temporary solution to tests with jest.

#57

PR #61, opened 2017-08-31
Adding field jsnext:main

matuzalemsteles:
For context #59.

### DIFF
--- a/packages/clay-button/webpack.config.js
+++ b/packages/clay-button/webpack.config.js
@@ -26,7 +26,7 @@ module.exports = {
 	},
 	plugins: [new webpack.optimize.ModuleConcatenationPlugin()],
 	resolve: {
-		mainFields: ['esnext:main', 'main'],
+		mainFields: ['esnext:main', 'main', 'jsnext:main'],
 		alias: {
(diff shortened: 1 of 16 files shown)


packages/clay-button/webpack.config.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
 	},
 	plugins: [new webpack.optimize.ModuleConcatenationPlugin()],
 	resolve: {
-		mainFields: ['esnext:main', 'main'],
+		mainFields: ['esnext:main', 'main', 'jsnext:main'],
```
jbalsas: Hey @matuzalemsteles, do we really need the `jsnext:main` main field here? I would simply keep the `esnext:main` if that's going to be the default future one. All the other ones can go. Same for the other `config` files.

jbalsas (dismissed):

packages/clay-button/webpack.config.js
matuzalemsteles: Hey @jbalsas,

I did some testing, we do not really need the `jsnext:main` field in `config` files. But we'll have to leave the `main` field because of the dependencies of some packages.

I'll send the change.

PR #62, opened 2017-09-04
LEXI-88 Create ClayBadge

matuzalemsteles:


brunobasto:
@matuzalemsteles can you put the ticket number on the commit message?

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,69 @@
...
+	{let $classes kind="text"}
+		badge badge-pill
+		{sp}badge-{$style ?: 'default'}
+
+		{if $size}
+			{sp}badge-{$size}
+		{/if}
+	{/let}
+
+	{let $attributes kind="attributes"}
```
brunobasto: This seems to only be necessary inside the `else` statement, right?

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Renders icon content
+ */
+{template .content}
+	{@param? closeable: bool}
+	{@param? label: html|string}
+	{@param? spritemap: string}
+
+	{$label ? $label : ''}
```
brunobasto: I don't see where this `label` param is being used. Do we need it?

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,69 @@
...
+
+	{if $closeable and $spritemap}
+		<button aria-label="Close" class="close" type="button">
```
brunobasto: How are we localizing hard-coded strings?

brunobasto (changes requested):

matuzalemsteles:
> can you put the ticket number on the commit message?

@brunobasto yes, do you think we should put it right in the `span` tag?

packages/clay-badge/src/ClayBadge.soy
matuzalemsteles: Yes, do you think we should put it right in the`span` tag?

packages/clay-badge/src/ClayBadge.soy
matuzalemsteles: Yes, when we do not pass `href` and we want to render only the `label` with icon or without.

packages/clay-badge/src/ClayBadge.soy
matuzalemsteles: We can locate by the metal selector, for example:
```javascript
events: {
	click: {
		selector: '.close',
		fn: handleClick
	}
}
```

matuzalemsteles:
Hey @brunobasto,

> can you put the ticket number on the commit message?

Yes, consider it done.

carloslancha:

carloslancha:
Ey @matuzalemsteles I'm reviewing [Lexicon's Badges documentation](https://docs.google.com/document/d/1ZU2UoH-MhGDH2fOsW-WyjLF3JD5iS2naMPI4KHDOd1U/edit) (attached in the ticket) and I've noticed that
- A badge cannot be a link
- A badge cannot be closeable

Don't know why I didn't see that reviewing your last pull, maybe I was looking in the wrong place. Sorry about that 😓 

If you have any doubt do not hesitate to ask me or Lexicon team

packages/clay-badge/src/ClayBadge.soy
brunobasto: Ah .Got it

packages/clay-badge/src/ClayBadge.soy
brunobasto: I meant `aria-label="Close"` will not be translated. Are there other components where we have hard-coded strings? If so, what's the pattern we are following?

carloslancha: Talking with @jbalsas we think we should go with it hardcoded for now while we find the best solution. It's an interesting issue...

brunobasto: Yeah :)

In portal, we have a mechanism in place that will translate keys used with
```soy
{msg desc=""}lang-key{/msg}
```

Not sure we could use it here, though

carloslancha: Yep, the point is that if we decide to go in that direction we should provide some "easy" mechanism to allow users to pass the keys to the component, or a series of .json with them that the build process can use... I don't know, just thinking loud... :P

carloslancha:
@matuzalemsteles I've talk with Victor Valle about the abbreviation stuff. This abbrs are not international, 100K could have no meaning for some cultures or 1M have different ones, so we decided to keep this as simple as possible and forget about the abbrs, leaving that decision to the user of the component. So Label should be a string.

Anyway, just as a hint for future developments, we need to have in mind both client and server renders. If we massage certain data in the JS (like here with the abbr) we have to replicate that behaviour in the soy in order to make both renders consistent.

Thx!

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -10,7 +10,7 @@
 
 	{let $classes kind="text"}
 		badge badge-pill
-		{sp}badge-{$style ?: 'default'}
+		{sp}badge-{$style}
```
carloslancha: If default value is `primary` as defined in the STATE this should be `{sp}badge-{$style ?: 'primary'}`

packages/clay-badge/src/ClayBadge.js
```diff
@@ -73,10 +30,10 @@ ClayBadge.STATE = {
 	 * The label of the badge element.
 	 * @instance
 	 * @memberof ClayBadge
-	 * @type {number}
+	 * @type {number|string}
 	 * default undefined
 	 */
-	label: Config.number().setter('hasNumberLabel_').required(),
+	label: Config.any().required(),
```
carloslancha: `Config.any()` accepts anything, including functions, arrays, etc. This should be just string.

carloslancha (changes requested):

packages/clay-badge/src/ClayBadge.js
```diff
@@ -0,0 +1,60 @@
...
+	 * @type {?string|undefined}
+	 * default undefined
+	 */
+	id: Config.string(),
+
+	/**
+	 * The label of the badge element.
+	 * @instance
+	 * @memberof ClayBadge
+	 * @type {string}
```
carloslancha: If required `@type` should be preceded by `!`

`@type {!string}`

packages/clay-badge/src/ClayBadge.js
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * Badge style color.
+	 * @instance
+	 * @memberof ClayBadge
+	 * @type {?string|undefined}
+	 * default undefined
+	 */
+	style: Config.oneOf([
+		'primary',
```
carloslancha: Order alphabetically

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,20 @@
+{namespace ClayBadge}
+
+/**
+ * This renders a badge.
+ */
+{template .render}
+	{@param? id: string}
```
carloslancha: Order params

packages/clay-badge/src/ClayBadge.soy
```diff
@@ -0,0 +1,20 @@
...
+
+	{let $classes kind="text"}
+		badge badge-pill
+		{sp}badge-{$style ?: 'primary'}
+	{/let}
+
+	<span class="{$classes}" id="{$id}">
+		{$label}
+	</span>
+
```
carloslancha: No empty line needed.

carloslancha (changes requested):
Could you please rebase with master and run `npm run mcritic` to confirm that passes validation?

Thx!

packages/clay-badge/src/ClayBadge.js
```diff
@@ -0,0 +1,60 @@
...
+	 */
+	label: Config.string().required(),
+
+	/**
+	 * Badge style color.
+	 * @instance
+	 * @memberof ClayBadge
+	 * @type {?string|undefined}
+	 * default undefined
+	 */
```
carloslancha, line 44: default is primary, not undefined

matuzalemsteles: You're right.

carloslancha:
LGTM Thx!!

PR #65, opened 2017-09-08
Add implicit params for mcritic and remove inherited params from component classes

carloslancha:


jbalsas:
LGTM!

@Robert-Frampton, are we fine relaxing the _Document all params even if inherited_? Makes sense to me...

PR #68, opened 2017-09-11
LEXI-89 Create ClayLabel

matuzalemsteles:

PR #76, opened 2017-09-13
Add alias `metal-soy` in webpack

matuzalemsteles:
Hey, guys,

A strange webpack behavior, when we depend on packets of clay, it happens that the webpack doubles, triples ... metal and other functions, depending on the amount of dependencies. The package becomes "giant" and `soy` when registering the models, it registers in different functions within that package by the same function in different places when executed.

```javascript
register(componentCtor, templates, mainTemplate = 'render') {
```

This implies running the ** javascript ** code inside the component. For a demonstration, see pr #68. 
Run a test by adding `console.log('--execute--') ` to` clay-icon` and try to execute packages that depend on it.

The solution to the problem was the addition of a `metal-soy` alias in the webpack.

### DIFF
--- a/packages/clay-badge/webpack.config.js
+++ b/packages/clay-badge/webpack.config.js
@@ -30,6 +30,7 @@ module.exports = {
 	resolve: {
 		mainFields: ['esnext:main', 'main'],
 		alias: {
+			'metal-soy': path.resolve('./node_modules/metal-soy'),
 			'incremental-dom': path.resolve('./node_modules/incremental-dom'),
 			'metal-incremental-dom': path.resolve(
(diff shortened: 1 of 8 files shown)


brunobasto:
Hey @jbalsas, we should probably try to automate this alias thing. It's getting bigger and I'm sure there are other modules getting duplicated on the final bundle that we are still not covering. ``metal-component`` for one.

This issue @matuzalemsteles fixed was interesting. We had duplicated definitions of the Soy template registry. So, each component had its own registry and they could not see each other. Even though nested components templates were rendered, the component constructors were never called.

What do you think?

PR #77, opened 2017-09-14
integrate axe-core wrapper to clay-buttons

vbence86:


jbalsas:
Just FYI for anyone reading, this is only a POC, not mergeable material yet 😉

jbalsas:
Hey @vbence86, tests are failing here...

vbence86:
This project is seemingly affacted by the `js-beautifier` issue reported [here](https://github.com/beautify-web/js-beautify/issues/1247).

https://travis-ci.org/metal/metal-clay-components/builds/276745452

```
npm ERR! enoent ENOENT: no such file or directory, chmod '/home/travis/build/metal/metal-clay-components/node_modules/js-beautify/js/bin/css-beautify.js' 
```

Do you want me to downgrade the version of `js-beautifier`?

jbalsas:
😢

We are bringing the dependency via tools, so we probably need to pin it down in there as well :(

vbence86:
I'm on it.

jbalsas:
hey @vbence86, I see rebase errors here... could you double check?

.babelrc
```diff
@@ -1,3 +1,4 @@
 {
-    "presets": ["metal"]
+    "presets": ["metal", "env"],
+    "plugins": ["syntax-async-functions", "transform-regenerator"]
```
jbalsas, line 3: Isn't this going to affect all builds we do? I assume this only needs to apply to the a11y bits, so maybe this actually demonstrates that we need to move that to another package...

jbalsas:
Hey @vbence86, why don't we set this up as another lerna package... `clay-a11y` so modules can depend on it and we have a simpler release and dep management story?

That way, we can also extract it later to its own repo...

vbence86:
sounds great to me, I felt the current solution a bit shoddy too.

vbence86:
The enhanced PR can be seen [here](https://github.com/metal/metal-clay-components/pull/79)

Issue #72, opened 2017-09-12
Transitive dependencies


carloslancha:
When we use a component as a dependency that uses another component as a dependency we need to add both as dependencies in our component in order to make soy compile working.

The problem is with `metalsoy --soyDeps`, as a workaround we made this:

https://github.com/metal/metal-clay-components/pull/71/commits/7319a9451cb303d8c51ba5dfd6da70f08d765bfd#diff-487d9b9c11d483b0c92610dbf8612825R21

matuzalemsteles:
Hey @carloslancha,

In the case of a component having two dependencies and both depend on another component, `metalsoy --soyDeps` accuses` Undefined template` and in case I add its example it accuses `Template 'ClayIcon.render' already defined`. Like the #68 case.

Do you have any insight into this?

julien:
I'll have a look into it.

carloslancha:
Fixed with https://github.com/metal/metal-clay-components/pull/82

PR #71, opened 2017-09-12
Create Clay-Navbar

carloslancha:

PR #78, opened 2017-09-18
Integrate yarn workspace with lerna

matuzalemsteles:
Hey guys,

I was working to solve the problem of jest solving that were found in pr #68.

The big problem was equal dependencies between different packages that caused jest and Webpack errors (in this case the number of alias started to grow, see pr #76), duplicating files within the bundle that caused various anomalies and not counting bundle was gigantic and resolution was resolved when it was pointed in one direction only.

I've been reading about [yarn workspace](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/) that has recently been released, its purpose is to facilitate the manipulation of package dependencies in monorepo repositories or with multi-package repositories, and [yarn came to version 1.0](https://github.com/yarnpkg/yarn/releases/tag/v1.0.0) on September 5th 😁.  

> Yarn Workspaces is a feature that allows users to install dependencies from multiple package.json files in subfolders of a single root package.json file, all in one go. 
> 
> Making Workspaces native to Yarn enables faster, lighter installation by preventing package duplication across Workspaces. Yarn can also create symlinks between Workspaces that depend on each other, and will ensure the consistency and correctness of all directories. [[more](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/)]

With this approach to yarn, it just fits nicely with our current clay architecture, as dependencies grow between packages, yarn is responsible for managing those dependencies in the root `node_modules`.

## Benefits

With this approach to yarn, we gain a number of benefits and solve various problems we are currently facing and solve possible problems in the future as new components emerge.

* We no longer need aliases in the packages Webpack df6abd2
* We remove several unnecessary codes in the solve jest 3ff677c
* Integrating simple with lerna
* We centralized equal dependencies between packages in the root folder
* Small bundle
* Our node_modules get much smaller

Our bundles were getting huge as the dependencies grew. Duplicating multiple files, such as metal, unnecessarily and causing side effects on some occasions, this is no longer the case in the yarn workspace approach. Let's for details, for example ClayLabel #68:

Dev - Webpack - Without yarn workspace
* Final compile: 1,3 MB
* Min+gz: 68.55 KB

Dev - Webpack - With yarn workspace
* Final compile: 618 KB
* Min+gz: 36.27 KB

## Not everything is flowers

The `metalsoy` package is not prepared for this type of architecture, we have some problems with this approach of yarn workspace, having all our clay components in the` node_modules` root, we can no longer pass the `--soyDeps` as something more generic and exit getting all the files with:

```
../../node_modules/clay-*/src/**/*.soy
```

This causes taking the component itself and in the compilation accuses that the component has already been declared, as the example below:

```
./../node_modules/clay-link/src/ClayLink.soy:6: error: Template 'ClayLink.render' already defined at src/ClayLink.soy:6:1
{template .render}
```

But we can work around this easily by adding to skip the component itself.

```
../../node_modules/clay-!(link)*/src/**/*.soy
```

But this still does not solve the problem, in the compilation another error returns:

```
../../node_modules/clay-navbar/src/ClayNavbar.soy:41: error: Illegal call to 'ClayLink.render', because according to the dependency graph, src/ClayLink.soy depends on ../../node_modules/clay-navbar/src/ClayNavbar.soy, not the other way around.
                {call ClayLink.render}
```

To function as expected we have to manage which dependencies we will pass to `metalsoy` in each package and if a package has no dependencies on other components of the clay, do not pass.

As an example, `clay-link` depends on` clay-icon`, thus:

```
../../node_modules/clay-icon/src/**/*.soy
```

But in specific cases like `clay-navbar` #71, we have dependence on `clay-link` that depends on `clay-icon`, the situation is that `metalsoy` does not understand that `clay-link` depends on `clay-icon` and resolve, see #72. In this situation depends on the must understand which dependencies, and add.

```
../../node_modules/clay-+(link|icon)/src/**/*.soy
```

The temporary resolution for this problem in all packages can be seen in the commit 3c1805137faf3c32ef3161a78cc7e90d728d9604.


jbalsas:
Hey @matuzalemsteles, this is an outstanding analysis, thanks so much for putting it together!! 👏 

I'm not super happy to move to a different package manager because we might lose sight of how others are using and consuming the library which could lead to us abandoning things such as the `jest-resolver` or `webpack` necessary configuration. We can't force everyone into using our tooling of choice.

That being said, if this does indeed solve our current problem in a simple way, it's hard for me to argue against it ☺️ 

@matuzalemsteles, tests are currently failing, so that's at least a must we need to fix before going any further
@Robert-Frampton, can you take a closer look at this approach for this repo? Also, since we're working on metal dev experience... how should we approach it in a general way? We're having trouble setting up testing and build environments for which we need both clear `webpack` and `jest-resolver` guidelines.

carloslancha:

lerna.json
```diff
@@ -1,7 +1,6 @@
 {
   "lerna": "2.0.0-rc.5",
-  "packages": [
-    "packages/*"
```
robframpton, line 4: Hey @matuzalemsteles 

Will removing this property affect the lerna publish task at all?

robframpton:
Hey guys,

I think that the issues we are running into are unique to a monorepo, which makes me comfortable with this approach because it's just a fix for a monorepo. It wouldn't be applicable to someone's app. In other words, our webpack config/tooling in a monorepo is going to be different than the webpack config/tooling a developer would have in an app. I don't think that we should worry about making them the exactly the same.

However, the concern over losing touch with how people are building these components is valid, but I think the solution is to develop some kind of integration testing that more closely emulates a production development environment. I'm not 100% sure what this would look like, but I think it's worth looking into.

My only other concern with this approach, is ensuring that all of the packages have the correct dependencies declared in their `package.json`. @matuzalemsteles maybe you can fill me in on how this works, but this is what I'm thinking.

Let's say we have two packages, and both implement some third party npm dependency, let's say `moment`.

```
packages/
  package-a (implements moment, declared in package.json)
  package-b (implements moment, NOT declared in package.json)
```

But now pretend that only `package-a` has `moment` listed as a dependency. Because all packages are now sharing a `node_modules` directory, `package-b` will technically work because it's using the `moment` installed from `package-a`. Once it's published though, `package-b` will break because it doesn't have `moment` installed.

Maybe there is some failsafe already in place that I'm not aware of, but if we can ensure that this never happens, I'm fine with this approach.

lerna.json
matuzalemsteles: Hey @Robert-Frampton,

When we enable yarn workspace in `lerna.json`:
```
{
  "npmClient": "yarn",
  "useWorkspaces": true
}
```

It will use the wire to initialize the project and will also use the `package.json/workspaces` field to find the packages instead of `lerna.json/packages`. I do not think this would have affected the publication.

matuzalemsteles:
@Robert-Frampton @jbalsas,

> Hey @matuzalemsteles, this is an outstanding analysis, thanks so much for putting it together!! 👏

Thanks 🙂!! 

> I'm not super happy to move to a different package manager because we might lose sight of how others are using and consuming the library which could lead to us abandoning things such as the jest-resolver or webpack necessary configuration. We can't force everyone into using our tooling of choice.

We will only be using the yarn workspace in our repo and when any developer uses the component, he can use `npm` quietly for his works. This approach will only apply to the monorepo and his administration. But it's important to think this way and agree with Robert about it.

> My only other concern with this approach, is ensuring that all of the packages have the correct dependencies declared in their package.json. @matuzalemsteles maybe you can fill me in on how this works, but this is what I'm thinking.

That thought is correct and we will have to make sure to keep all the dependencies explicit in the `package.json` of each package. [In that post](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/), you have a clearer example of this.

carloslancha:

carloslancha:
Resent https://github.com/metal/metal-clay-components/pull/82

jbalsas:
#82 Was merged!! 🎉

PR #79, opened 2017-09-19
add clay-a11y as dependency across all packages

vbence86:
### Disclaimer
Accessibility violations are not detected and therefore not captured that has a negative impact on the output of the website using clay components.

### Issue
Accessibility linters in existence are predominantly employ **Static analysis** and fail to check HTML output placed into template files such as **JSX** or **SOY** due to dynamic parts and attributes cannot be fully evaluated without compilation. 

### Solution
Added a self-contained tool that with the power of the popular `axe-core` module checks the component exactly as it is going to be rendered into the consuming service.

#### How it works
1. It fires up an `express` server against the demo assets the individual clay component incorporates.
2. Creates an instance of headless Chrome environment through ``puppeteer``.
3. Loads the specified page.
4. Injects `axe-core` to the demo page and executes the accessibility check to collect potential violations.
5. Outputs a report on possible accessibility violations.

### Usage
The tool has been added as a `lerna` package. To activate it in a clay component it needs to be provided as a dependency in `package.json`. 

##### Add to the package
```
"devDependency": { 
  ...
  "clay-a11y": "^1.0.0",
  ...
}
```

##### Execute 
```
"script": {
  ...
  "a11y": "npm run build && clay-a11y -c demos/index.html"
  ...
}
```

clay-a11y


packages/clay-a11y/README.md
```diff
@@ -0,0 +1,26 @@
+# clay-badge
+
+Clay Badge
```
jbalsas: Wrong readme ;)

package.json
```diff
@@ -12,6 +12,7 @@
   },
   "devDependencies": {
     "babel-preset-metal": "^4.1.0",
+    "clay-lerna-jest-resolver": "^1.0.1",
```
jbalsas: Do we need to add this?

packages/clay-a11y/bin/clay-a11y
```diff
@@ -0,0 +1,13 @@
...
+SCRIPT_PATH="${BASH_SOURCE[0]}";
+while([ -h "${SCRIPT_PATH}" ]); do
+    cd "`dirname "${SCRIPT_PATH}"`"
+    SCRIPT_PATH="$(readlink "`basename "${SCRIPT_PATH}"`")";
+done
+cd "`dirname "${SCRIPT_PATH}"`" > /dev/null
+SCRIPT_PATH="`pwd`";
+popd  > /dev/null
+
+${SCRIPT_PATH}/../node_modules/.bin/babel-node ${SCRIPT_PATH}/../src/axe-runner.js "$@"
```
jbalsas, line 13: Is this windows-safe?

packages/clay-a11y/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "clay-a11y",
+  "version": "1.0.0",
+  "description": "Accessibility test module for clay components",
+  "main": "index.js",
+  "bin": {
+    "clay-a11y": "./bin/clay-a11y"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
jbalsas: We can probably get rid of this ;)

jbalsas (changes requested):
Hey @vbence86 this looks very very promising!!

Currently, the CI is failing with a `TypeError: chalk.stripColor is not a function` error. Could you take a look?

Other than that, I just added some minor nits :)

packages/clay-a11y/bin/clay-a11y
vbence86: I'll have to look into it. It's purpose is to provide a way to use babel-node instead of node.

vbence86: After spending a bit of time researching how easy it might be to make cross platform `npm scripts`, I'm confident to say it is a great deal of effort in this case. Furthermore, Windows 10 now supports bash scripts so there is no complete lack of support on Windows side for this script. 

Alternative solutions:
##### Use purely supported ES5/ES6 language features
- Remove any syntax that is not currently supported by `Node 8`
- Execute the a11y script with Node instead of `babel-node`
- Remove the wrapper bash script

##### Compile stand-alone executable 
- compile a binary with the required version of Node bundle into it (Nexe)

jbalsas:
We're trying to extract the a11y module so we can use it properly from the root level, so closing here!

PR #82, opened 2017-09-20
Make soyDeps more explicit

carloslancha:


jbalsas:
Based on the comments in https://github.com/metal/metal-clay-components/pull/78, I'm going to merge this in to keep going.

The only open issue we had was the one @Robert-Frampton brought up about hidden dependencies:
> Maybe there is some failsafe already in place that I'm not aware of, but if we can ensure that this never happens, I'm fine with this approach.

I'm not sure how we can enforce this... one thing we noticed was that we were being too lenient with our `soyDeps`, adding more than really necessary, hiding the real dependencies. On another note, transitive dependencies resolved by `npm` are not picked up by `soyDeps`, so there's more to do there...

I guess this could somehow be addressed based on:
>[...] I think the solution is to develop some kind of integration testing that more closely emulates a production development environment. I'm not 100% sure what this would look like, but I think it's worth looking into.

🎉

jbalsas:
Thanks @matuzalemsteles, @Robert-Frampton and @carloslancha for working this out! 👏

brunobasto:
Awesome!

PR #85, opened 2017-09-21
Update ClayCheckbox markup

carloslancha:


jbalsas:
Hey @carloslancha, why the API change here?

carloslancha:
@jbalsas trying to be clearer. We were not hiding but not printing the wrapper in that case. So, since the mission of the wrapper is to make input to be display as a block, I think this API describe it better.

PR #88, opened 2017-09-21
Applies new formatting

jbalsas:
Hey @carloslancha, @Robert-Frampton, @brunobasto, @matuzalemsteles, this is (hopefully) our last guidelines change ☺️ 

I've applied `prettier-eslint` so we consolidate all the guidelines in [eslint-config-liferay](https://github.com/eduardolundgren/eslint-config-liferay).

What do you think?

/cc @eduardolundgren


matuzalemsteles:
Hey @jbalsas,

Sounds good.

Just to clarify a point, the recommended would be to use `yarn add prettier-eslint-cli -D` to add new packages, to let yarn take care of dependencies and end up not messing up the tree of dependencies. When adding a dependency on a particular package it is recommended that you use it as well.

When run inside the package, yarn is responsible for administering the dependency and updating the `yarn.lock` in root. Using npm it does not understand this tree of dependencies of the yarn and can end up messing up.

jbalsas:
Right, @matuzalemsteles, I forgot we were using yarn now 😂 

Just to clarify, this is a root-level dependency only, so there's no need to add them in the independent packages.

I'll push an extra commit updating the `yarn.lock`

jbalsas:
Since no one has complained, I'll merge and we can adapt later 🤷‍♂️

PR #89, opened 2017-09-22
LEXI-153 ClayAlert

carloslancha:


matuzalemsteles:

packages/clay-alert/src/ClayAlert.js
```diff
@@ -0,0 +1,203 @@
...
+ */
+ClayAlert.STATE = {
+  /**
+   * Flag to indicate if alert should be automatically closed.
+   * @instance
+   * @memberof ClayAlert
+   * @type {?bool}
+   * @default undefined
+   */
+  autoClose: Config.bool(),
```
matuzalemsteles: @carloslancha, would not it be better to set `boolean` values with a default? `Config.bool().value(false)` add this as practice, what do you think?

matuzalemsteles:
@carloslancha,

I think it would be unnecessary at the moment to add `package-lock.json` to the packages.

packages/clay-alert/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "keywords": [
+    "clay",
+    "metal"
+  ],
+  "dependencies": {
+    "clay-button": "^0.0.0",
+    "clay-icon": "^0.0.0",
+    "metal": "^2.13.2",
+    "metal-component": "^2.13.2",
+    "metal-dom": "^2.13.2",
```
matuzalemsteles: I think it would be better to remove from here, since it is not being used.

packages/clay-alert/src/ClayAlert.soy
```diff
@@ -0,0 +1,80 @@
...
+					'monospaced': false,
+					'spritemap': $spritemap,
+					'symbol': 'times'
+					] /}
+				{param style: 'unstyled' /}
+				{param type: 'button' /}
+			{/call}
+		{/if}
+
+		{let $icon kind="text"}
```
matuzalemsteles, line 56: This variable below the `attributes` variable. Would not it be more comfortable reading the rest of the code? since, it is not being influenced by the position. what do you think?

carloslancha: Until now we're defining variables right before being used. Maybe we can discuss it to add it to the guidelines. Thx for bringing it up!

packages/clay-alert/src/ClayAlert.js
carloslancha: Makes sense!

matuzalemsteles:
LGTM!

PR #97, opened 2017-09-26
Feature/a11y

vbence86:
#### Intent
To integrate accessibility test to metal-clay-components.


carloslancha:
Don't forget to add `role="radiogroup"` to the `ClayRadio` wrapper in https://github.com/metal/metal-clay-components/pull/97/commits/181f064bcb978cb006000afcb270eda2fb50ca26#diff-51e59515af5c50ecfe051746bc4b243bR29 ;)

vbence86:
@carloslancha that's the thing, it needs to be added to a container element that incorporates all elements with the same `name` attribute. This needs to be controlled either in the Demo or a prospective clay component that is responsible for containing ClayRadios.

carloslancha:
@vbence86 Ah ok! It's only when radios share the same name! Isn't it? I didn't get it right before, sorry? 😅

packages/clay-button/src/ClayButton.soy
```diff
@@ -135,8 +129,12 @@
 
 	{if $ariaLabel}
 		aria-label="{$ariaLabel}"
+	{elseif $label}
```
carloslancha, line 141: You need to add `label` and `icon` as parameters of `.attributes` template and pass'em from `.render` (That's why compilation is failing)

vbence86: Yes, I got a failing build and I've amended this by now, thanks for the head-ups though.

vbence86:
Yes, individual radio buttons with unique `name` attribute pass a11y with no need of any additional aria tags.

vbence86:
a11y module looks good now. We've got few violations captured with regards to the newly added ClayDropDown. I'm fixing them now.

captura de pantalla 2017-09-27 a las 9 23 20

jbalsas:
@matuzalemsteles, @carloslancha, can you review the changes here?

carloslancha:

carloslancha:
LGTM! Good job!!

packages/clay-dropdown/src/ClayDropdownBase.soy
```diff
@@ -135,7 +134,7 @@
 		"
 	{/let}
 
-	<ul {$listAttributes}>
+	<div {$listAttributes}>
```
jbalsas: why are we moving away from semantic tags?

vbence86: `<ul>` elements on no account must incorporate elements other than `<li>`, `<script>`, `<template>`. (reference - https://dequeuniversity.com/rules/axe/1.1/list)

A solution would be to append all content wrapped by `<li>` elements, but due the code there seems to wrap everything in `<div>` I thought I'd follow that approach and remove `<ul>` completely. This change could likely imply unexpected design consequences though.

jbalsas: If this is a list of items, it needs to be a list. If an item is not an item, either it needs to be put outside the list or the list needs to be declared deeper down to enclose the real list of items.

vbence86: Agreed. With this in mind, I will have to refactor the whole dropdown soy file.

jbalsas: 👍

carloslancha: I think making all `{deltemplate ClayDropdown.item}` and variants `li` elements (and `ul` the parent) and adding `aria-hidden` and `role=presentation` to the `separator` and `header` variants should do the work.

vbence86: I'm addressing this.

PR #102, opened 2017-09-28
LEXI-162 ClayDropdown

carloslancha:


carloslancha:
@matuzalemsteles could you please take a look at this? Thx!

matuzalemsteles:

packages/clay-dropdown/src/ClayDropdown.js
```diff
@@ -89,11 +90,20 @@ ClayDropdown.STATE = {
...
 	 * @type {!string}
 	 * @default unstyled
 	 */
   triggerStyle: Config.oneOf(['secondary', 'unstyled']).value('unstyled'),
+
+  /**
+   * Type of the dropdown menu.
+   * @instance
+   * @memberof ClayDropdown
+   * @type {!string}
```
matuzalemsteles: Not required. We can add `{?string}`.

packages/clay-dropdown/src/ClayDropdown.js
```diff
@@ -89,11 +90,20 @@ ClayDropdown.STATE = {
...
+   * @default list
+   */
+  type: Config.oneOf(['form', 'list']),
```
matuzalemsteles: Since the default is `list`, we can add `Config.oneOf(['form', 'list']).value('list')`. What do you think?

packages/clay-dropdown/src/ClayDropdown.js
```diff
@@ -89,11 +90,20 @@ ClayDropdown.STATE = {
   /**
 	 * Style of the trigger button.
 	 * @instance
-	 * @memberof ClayDropdownBase
+	 * @memberof ClayDropdown
 	 * @type {!string}
```
matuzalemsteles: As the default is `unstyled` I see no sense in being able to say that it is required. Add as optional?

packages/clay-dropdown/src/ClayDropdown.js
```diff
@@ -21,13 +21,14 @@ ClayDropdown.STATE = {
   /**
    * Button configuration to place a button at dropdown footer.
    * @instance
-   * @memberof ClayDropdownBase
+   * @memberof ClayDropdown
    * @type {?string}
    * @default .dropdown-toggle
```
matuzalemsteles: Where would the `.dropdown-toggle`?

packages/clay-dropdown/src/ClayDropdownBase.js
```diff
@@ -72,6 +72,7 @@ ClayDropdownBase.STATE = {
   button: Config.shapeOf({
     label: Config.string().required(),
     style: Config.oneOf(['primary', 'secondary']).value('primary'),
+    type: Config.oneOf(['button', 'reset', 'submit']),
```
matuzalemsteles: What do you think of adding `oneOf` default values as a practice?

packages/clay-dropdown/src/ClayDropdownBase.js
```diff
@@ -162,6 +163,15 @@ ClayDropdownBase.STATE = {
...
   ]).value('unstyled'),
+
+  /**
+   * Type of the dropdown menu.
+   * @instance
+   * @memberof ClayDropdownBase
+   * @type {!string}
+   * @default list
+   */
+  type: Config.oneOf(['form', 'list']),
```
matuzalemsteles: Add the default value `value('list')`.

packages/clay-dropdown/src/ClayDropdownBase.js
```diff
@@ -162,6 +163,15 @@ ClayDropdownBase.STATE = {
     'secondary',
     'unstyled',
   ]).value('unstyled'),
+
+  /**
+   * Type of the dropdown menu.
+   * @instance
+   * @memberof ClayDropdownBase
+   * @type {!string}
```
matuzalemsteles: Would it be optional or required? In `soy` is it optional.

packages/clay-dropdown/src/ClayActionsDropdown.js
```diff
@@ -19,6 +19,28 @@ class ClayActionsDropdown extends Component {}
  */
 ClayActionsDropdown.STATE = {
   /**
+   * Button configuration to place a button at dropdown footer.
+   * @instance
+   * @memberof ClayDropdown
+   * @type {?string}
+   * @default .dropdown-toggle
```
matuzalemsteles: Where would the `.dropdown-toggle`?

packages/clay-dropdown/src/ClayDropdown.js
carloslancha: Copy/paste problems 😅

packages/clay-dropdown/src/ClayDropdown.js
carloslancha: Is not required, wrong `!` in `@type`

packages/clay-dropdown/src/ClayDropdown.js
carloslancha: Is not required, wrong ! in @type

packages/clay-dropdown/src/ClayDropdown.js
carloslancha: Sure

packages/clay-dropdown/src/ClayDropdownBase.js
carloslancha: Is not required, wrong ! in @type

packages/clay-dropdown/src/ClayDropdownBase.js
carloslancha: Depends on if we want it to have a default value or not. In this case default can be `button`, but there can be cases where we want default to be undefined.

carloslancha: In the case of the `button` param in `ClayDropdown` I wouldn't add defaults since we're passing entire object to `ClayDropdownBase` and we can't define that default in soy, so is `ClayDropdownBase` the one responsible of that.

jbalsas:
Thanks @carloslancha! @matuzalemsteles, I'm merging this in, but feel free to suggest any additional changes and we can introduce them later on!

PR #110, opened 2017-10-09
Fix duplication of the a11y in ci

matuzalemsteles:
Was running `a11y` twice on the ci because the `test` script was already running `a11y`.


carloslancha:
LGTM

PR #111, opened 2017-10-10
Replace babel-preset-es2015 and update the version of lerna and yarn

matuzalemsteles:
* Replace **babel-preset-env** with babel-preset-env. [more](https://babeljs.io/env/)...
* Yarn [1.2.0](https://github.com/yarnpkg/yarn/releases/tag/v1.2.0)
* Lerna [2.4.0](https://github.com/lerna/lerna/releases/tag/v2.4.0)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ before_install:
   - nvm install 8
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.1.0
   - export PATH=$HOME/.yarn/bin:$PATH
-  - npm install -g lerna@2.2.0
+  - npm install -g lerna@2.4.0
 
 cache:
   yarn: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 before_install:
   - nvm install 8
-  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.1.0
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.2.0
   - export PATH=$HOME/.yarn/bin:$PATH
   - npm install -g lerna@2.4.0
 
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.2.0",
+  "lerna": "2.4.0",
   "version": "1.0.0-alpha.1",
   "npmClient": "yarn",
   "useWorkspaces": true
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "http-server": "^0.9.0",
     "husky": "^0.14.2",
     "jest": "^20.0.4",
-    "lerna": "^2.2.0",
+    "lerna": "^2.4.0",
     "lint-staged": "^4.0.0",
     "metal-a11y-checker": "^1.0.5",
     "metal-jest-serializer": "^1.0.1",
--- a/packages/browserslist-config-clay-components/README.md
+++ b/packages/browserslist-config-clay-components/README.md
@@ -3,7 +3,7 @@ Clay components shareable config for Browserslist.
 
 ## Installation
 ```
-npm install browserslist-config-clay-components
+npm install browserslist-config-clay-components -D
 ```
 
 ## Usage
--- a/packages/clay-alert/package.json
+++ b/packages/clay-alert/package.json
@@ -41,5 +41,8 @@
     "clay": "liferay/clay#develop",
     "metal-tools-soy": "^4.2.1",
     "webpack": "^3.0.0"
-  }
+  },
+  "browserslist": [
+    "extends browserslist-config-clay-components"
+  ]
 }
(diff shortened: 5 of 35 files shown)


jbalsas (commented):
Hey @matuzalemsteles, these changes look great, I just have a doubt... what is the default browserlist that the preset accounts for?

We should have `all latest, ie11`. Could you double check to make sure we're not accidentally skipping IE11?

matuzalemsteles:
Hey @jbalsas,

It may seem that `babel-preset-env` behaves the same as babel-preset-latest (or babel-preset-es2015, babel-preset-es2016, and babel-preset-es2017 together), we can assure ourselves of supporting IE11 , using:
```javascript
presets: [['babel-preset-env', {
    "targets": {
        "browsers": ["last 2 versions"]
     }
}]],
```

Until check if everything is okay, but by default supports IE10 and IE11 as default. What do you think?

robframpton:
Hey @jbalsas 

There is a web app that let's you test browserlist patterns, we might be able to get away with `last 1 version`. Though I'm not sure we should drop IE10, since DXP currently supports it.

http://browserl.ist/?q=last+1+version

Or if we need older versions of Safari for example.

http://browserl.ist/?q=last+1+version%2C+safari+9-10

matuzalemsteles:
Hey @Robert-Frampton,

From what I was seeing, `babel-preset-env` uses the` browserlist` and by default `browserlist` supports IE10 and IE11.

jbalsas:
Hey @Robert-Frampton, @matuzalemsteles, `clay-components` are going to be released with `Liferay Portal 7.1`, which will only support `IE11`, so there's no need for us to support `<IE11` in this project. If that can save us some transpiling, let's welcome it ;)

jbalsas:
@matuzalemsteles, can we set the `env` settings to make sure we're only targetting `IE11` from the `IE` brand?

matuzalemsteles:
Hey @jbalsas, That looks great 😆, we can use `> 1%` to ensure good coverage and we will be reaching `IE 11`, but if you want to make sure we can use `> 1%, IE 11`. What do you think ? http://browserl.ist/?q=%3E+1%25%2C+IE+11

jbalsas:
Will we be adding `Samsung Internet4` and `Chrome49` with that `>1%`? Because I don't think we want that 😂

jbalsas:
I think [last 1 version](http://browserl.ist/?q=last+1+version) should be fine... not sure how `preset-env` targets mobile browsers though...

matuzalemsteles:
😂 Okay, I think we can improve `> 1.5%, Opera > 46, Safari > 10`, http://browserl.ist/?q=%3E+1.5%25%2C+Opera+%3E+46%2C+Safari+%3E+10. what do yout think?

jbalsas:
Hey @matuzalemsteles, we can't rely on market share, really. This is roughly our support matrix directive:

`Last version of every browser that has automatic updates plus IE11` 😉

jbalsas:
On top of that, we also usually support ESR or LTS versions, like [Firefox ESR](https://www.mozilla.org/en-US/firefox/organizations/), so we might want to throw that one as well.

matuzalemsteles:
You're right, let's use`last 1 version` ?

jbalsas:
Yeah, let's go with `last 1 version, Firefox ESR` (we need to translate this). Could you also verify before what that means for mobile browsers? Because if the preset takes them into account, we will likely never get rid of some transpilation if mobile browsers aren't updated... can we set a specific query to filter out most mobile browsers?

matuzalemsteles:
We do not have a specific query to remove mobile browsers, we will have to be less generic in our queries.

About `Firefox ESR` is only available for desktop, to make sure 😀:
https://wiki.mozilla.org/Enterprise/Firefox/ExtendedSupport:Proposal#Caveats
https://www.mozilla.org/en-US/firefox/organizations/faq/

jbalsas:
What about [last 2 Chrome versions, last 1 Edge versions, last 2 Firefox versions, Firefox 52, IE 11, last 1 Opera versions, last 2 Safari versions, last 1 ChromeAndroid versions, last 2 iOS versions](http://browserl.ist/?q=last+2+Chrome+versions%2C+last+1+Edge+versions%2C+last+2+Firefox+versions%2C+Firefox+52%2C+IE+11%2C+last+1+Opera+versions%2C+last+2+Safari+versions%2C+last+1+ChromeAndroid+versions%2C+last+2+iOS+versions)?

matuzalemsteles:
Sounds great to me. What do you think about centralizing this list in a package? something like `browserslist-config-clay` and add in this repository, so we can use:
```
"browserslist": [
    "extends browserslist-config-clay"
]
```

jbalsas:
Sounds good, @matuzalemsteles! Can you put one in here for now, let's make it `browserslist-config-clay-components` so we can start referencing in this PR.

Once we have this package inside the `metal-clay-components` repo we can release it and start using it here!

I think we'll probably want to create a liferay generic one, so we can apply it in other components as well, but I can take care of that over the week.

matuzalemsteles:
Sounds great, I'll push that now to that repo, so we can start using it in `metal-clay-components`.

matuzalemsteles:
#112

jbalsas:
Hey @matuzalemsteles, for some reason the Travis status is not getting back, so I can't really merge without it since `master` is a protected branch... Could you just resend this so we can see the tests run once more?

Thanks!

PR #119, opened 2017-10-19
LEXI-64 Create ClayModal

matuzalemsteles:
### Animations
For the correct functioning of `ClayModal` with animations, it was necessary to add a `timeout` to create a delay and add the `show` class to the` overlay` after it is put into `dom`, the same happens when showing the modal , must add a `_isTransitioning` variable to add the` show` class to the modal. Can not add all at the same time when are injected into the dom, because `display: block` nullifies the transition.

```javascript
show() {
  dom.addClasses(document.body, 'modal-open');
  this.visible = true;

  setTimeout(() => {
      this._isTransitioning = true;
      dom.addClasses(this._overlayElement, 'show');
   }, 5);

...
```
```soy
{let $attributes kind="attributes"}
    class="
	fade modal
	{if $_isTransitioning}
	    {sp}show
	{/if}
    "

    {if $visible}
	style="display: block;"
    {/if}
...
```

### Tests
To test when the modal is opened or closed, it was necessary to add `timeout` to wait for the animations to be made and reflected in` dom` and thus to complete the test.

### DIFF
--- a/packages/clay-modal/LICENSE.md
+++ b/packages/clay-modal/LICENSE.md
@@ -1,0 +1,29 @@
+# Software License Agreement (BSD License)
+
+Copyright (c) 2014, Liferay Inc.
+All rights reserved.
+
+Redistribution and use of this software in source and binary forms, with or without modification, are
+permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above
+  copyright notice, this list of conditions and the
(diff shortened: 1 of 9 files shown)


carloslancha:

carloslancha:
Resent https://github.com/metal/metal-clay-components/pull/125

PR #125, opened 2017-10-25
LEXI-64 ClayModal

carloslancha:


carloslancha:
Hey @matuzalemsteles ! I've done some changes in the modal to adjust it better to the markup and to get rid of those setTimeouts. What do you think? :)

matuzalemsteles:

packages/clay-modal/src/ClayModal.js
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  /**
+   * @inheritDoc
+   */
+  // eslint-disable-next-line
+  sync_isTransitioning() {
+    if (this._isTransitioning && !this.visible) {
+      this._isTransitioning = false;
+      this.visible = true;
```
matuzalemsteles, line 63: Hey @carloslancha,

One of my concerns is when the change happens at the same time this takes the animation out of the modal, sometimes this change time is inconsistent and appears to be a bug when it does not have the animation, I would not want to trust it. What do you think ?

packages/clay-modal/webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+        use: {
+          loader: 'babel-loader',
+          options: {
+            compact: false,
+            presets: [
+              [
+                'babel-preset-env',
+                {
+                  targets: {
+                    browsers: ['last 2 versions'],
```
matuzalemsteles, line 20: It is no longer necessary to use it since we are extending it from `browserslist-config-clay-components`. I forgot to remove 😅

matuzalemsteles:
Hey @carloslancha, Much better now, thanks for the tweaks.

What do you think to create a test to cover the use of `show ()`?

packages/clay-modal/src/ClayModal.js
carloslancha: There's always a render between syncs so will not happen at the same time. This is the only place where both params are set together, but in this case is just to "reset" `isTransitioning` and has no effect on render logic :)

carloslancha:
Hey @matuzalemsteles sure, we need to cover that method, but we definitely should not use timeouts on tests, I hadn't time today to work on that but I wanted to send it because I'll be on vacation till next thursday. Could you work on it if you have time?

Thx and good job!

carloslancha:
Another thing I thought about is that maybe we should have a `hide` method that calls `hide` event, and work with `show` method in the same way we're doing with `hide`, by emitting a `show` event and calling a `_defaultShowModal` or something like that. What do you think @matuzalemsteles ?

packages/clay-modal/webpack.config.js
carloslancha: Hahaha, I didn't see it 😅

packages/clay-modal/src/ClayModal.js
matuzalemsteles: Hey @carloslancha, In some tests sometimes Modal had no effect. Only guarantee the consistency of the `overlay`, I will see this and inform you better.

matuzalemsteles:
> Hey matuzalemsteles sure, we need to cover that method, but we definitely should not use timeouts on tests, I hadn't time today to work on that but I wanted to send it because I'll be on vacation till next thursday. Could you work on it if you have time?

Okay, I'll take a look at this. By the way, a good vacation.

> Another thing I thought about is that maybe we should have a hide method that calls hide event, and work with show method in the same way we're doing with hide, by emitting a show event and calling a _defaultShowModal or something like that. What do you think matuzalemsteles ?

Look good to me, we can put a `show` event in `attached` just like `hide` and do that. I'll work on it. Thanks!

matuzalemsteles:
Resent #137

PR #134, opened 2017-10-30
Update ClayAlert markup

### LINKED ISSUE #127: ClayAlert markup
We need to update markup to clay 2.0.0 Beta.

matuzalemsteles:
This resolves #127.

### DIFF
--- a/packages/clay-alert/src/ClayAlert.soy
+++ b/packages/clay-alert/src/ClayAlert.soy
@@ -36,27 +36,24 @@
 		role="alert"
 	{/let}
 
+	{let $content kind="html"}
+		{call .content}
+			{param _handleCloseClick: $_handleCloseClick /}
+			{param closeable: $closeable /}
+			{param message: $message /}
+			{param spritemap: $spritemap /}
+			{param style: $style /}
+			{param title: $title /}
+		{/call}
+	{/let}
+
 	<div {$attributes}>
 		{if $type == 'fluid'}
 			<div class="container">
-				{call .content}
-					{param _handleCloseClick: $_handleCloseClick /}
-					{param closeable: $closeable /}
-					{param message: $message /}
-					{param spritemap: $spritemap /}
-					{param style: $style /}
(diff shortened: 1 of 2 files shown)


carloslancha:

packages/clay-alert/src/ClayAlert.soy
```diff
@@ -37,43 +37,79 @@
...
-				{param icon: [
-					'alignment': 'right',
-					'spritemap': $spritemap,
-					'symbol': 'times'
-					] /}
-				{param style: 'unstyled' /}
-				{param type: 'button' /}
+		{if $type == 'fluid'}
+			<div class="container">
+				{call .content}
```
carloslancha: Due is the same content I'd say to extract this `call` out of the `{if}` and create a variable `$content`

```
	{let $content kind="html"}
		{call .content}
			{param _handleCloseClick: $_handleCloseClick /}
			{param closeable: $closeable /}
			{param message: $message /}
			{param spritemap: $spritemap /}
			{param style: $style /}
			{param title: $title /}
		{/call}
	{/let}

	{if $type == 'fluid'}
		<div class="container">
			{$content}
		</div>
	{else}
		{$content}
	{/if}
```

carloslancha (changes requested):
We need to add to the markup the notification type too. Be sure that notifications can be stacked :)

matuzalemsteles:
Hey @carloslancha,

Our current ClayAlert API would not support the use of the `alert-notifications-fixed` or `alert-notifications-absolute` notification type because we treat Alert as only a rendered component, not multiple. The current marking of these types is to wrap the alerts within `alert-notifications-fixed` and` alert-notifications-absolute`, for this we will have to treat ClayAlert as several alerts. What do you think about this?

carloslancha:
I see... we need to think about this carefully. Let's merge this one and open an issue for the notifications part :)

Thx!

carloslancha (approved):

PR #147, opened 2017-11-15
IFI-115 Create placeholder feature for SSR charts

julien:


jbalsas:
@Robert-Frampton, if you have the chance to take a quick look at these, since you worked on the original implementation... @carloslancha and @matuzalemsteles can start taking over starting here with some help ;)

packages/clay-charts/src/AreaLineChart.soy
```diff
@@ -4,5 +4,26 @@
...
+		{let $placeholderAttrs kind="attributes"}
+			{if not $loading}
+				hidden
+			{/if}
+		{/let}
+
+		<div ref="placeholder" {$placeholderAttrs}>
+			{call Chart.loading data="all" /}
+		</div>
+	</div>
```
robframpton: Hey @julien, I think it should be possible to pull this markup out into a partial of some kind. It would be nice to not need to duplicate.

robframpton (changes requested):

packages/clay-charts/src/AreaLineChart.soy
julien: @Robert-Frampton of course. I'll do that.

packages/clay-charts/src/DonutChart.soy
```diff
@@ -4,5 +4,7 @@
  * This renders the component's whole content.
  */
 {template .render}
-	<div class="metal-chart"></div>
+	{@param? loading: bool}
+
+	{call Chart.chart data="all" /}
```
carloslancha: @julien we should avoid the use of `data="all"` and try to explicitly pass the params.

carloslancha (changes requested):

packages/clay-charts/src/DonutChart.soy
carloslancha: https://hosting-liferayfrontendguidelines.wedeploy.io/soy/conventions.html

julien:
@carloslancha thanks for the help.
Let's see if there's something we can do about`mcritic` errors ;)

jbalsas:
Hey @julien, let's just use the `implicitParams` config of mcritic for now!

julien:
@jbalsas ok thanks.

carloslancha:
Merged, thx! :)

PR #152, opened 2017-11-17
Removes Mocha in favor of Jest for generator testing

zenorocha:


jbalsas:
Hey @zenorocha, I think this should be fine, just not the renaming of the `__tests__` since that's actually the jest default: https://facebook.github.io/jest/docs/en/configuration.html#testmatch-array-string

zenorocha:
I'm actually only renaming the generator's input folder, not the output: https://github.com/metal/metal-clay-components/pull/152/files#diff-d7f0e8e825e636b8ed5b35166aa43553R100

This is needed because jest will try to test that folder as well.

jbalsas:
Ah, that makes sense! 😂 

We're also missing the `eslint` config and some default scripts in there... I'll open an issue for that!

PR #153, opened 2017-11-17
Fixes repository definition on package.json

zenorocha:


jbalsas:
Hey @zenorocha, I'm not sure about this change... almost every project I've seen points to the repo, not to the independent packages... this field is meant to be consumed by computers, for example to clone something, so I'm not sure if those url would work...

I've just seen babel use this format, I think... Are you aware of any specific literature or recommendations for this? How is this better?

zenorocha:
I've actually seen this on https://www.npmjs.com/package/metal-events. The reason why I liked it's because you from easily click from npm to github without navigating through the folders. But I'm ok with not using it, just found it convenient.

jbalsas:
Yeah, that's the thing I also find convenient, but wondered why noone else is doing it...

From [package.json|repository](https://docs.npmjs.com/files/package.json#repository)

> Specify the place where your code lives. This is helpful for people who want to contribute. If the git repo is on GitHub, then the npm docs command will be able to find you.

> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It's for computers.

The "_it's for computers_" part is what confuses me, but if we don't know exactly what that means, I'm fine merging this until we know more 😂 

The more common configuration for this I've seen on lerna repos is:

```json
"repository": {
  "type": "git",
  "url": "https://github.com/facebook/jest.git"
}
```

Babel is a big project that uses the [specific package url](https://github.com/babel/babel/blob/master/packages/babel-core/package.json#L9), though 😉

zenorocha:
Nice, thanks for checking that out. My fine with either one ;)

jbalsas:
@Robert-Frampton, do you have any insight on this? 🤷‍♂️

robframpton:
Not really, the only application of the url that really matters to me is the link on the packages npm page. As long as that link works I'm fine with either.

jbalsas:
Pull request submitted to https://github.com/metal/metal-clay-components/pull/164. <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

jbalsas:
Hey @zenorocha, since we haven't got any answer, I've resolved the conflict and resend as https://github.com/metal/metal-clay-components/pull/164. I will just merge and we can rethink it if we ever gain knowledge of a better way to handle this.

zenorocha:
Thanks @jbalsas!

PR #158, opened 2017-11-20
ClayAlert types fix

### LINKED ISSUE #154: ClayAlert type 'fluid'
Based on [Lexicon Alert Pattern](https://lexicondesign.io/docs/patterns/alerts.html), the `fluid` type of alert is actually called `stripe`. For simpler usage and mapping, we should match Lexicon's type.

Also, 

> The close action is mandatory in this alert type.

So it shouldn't be necessary to set closeable to true for these type of alerts.

carloslancha:
This resolves #154.

### DIFF
--- a/packages/clay-alert/demos/index.html
+++ b/packages/clay-alert/demos/index.html
@@ -70,23 +70,23 @@
 
 	<div class="row row-spacing">
 		<div class="col">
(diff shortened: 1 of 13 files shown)


matuzalemsteles:

matuzalemsteles:
Merged, thanks!

PR #167, opened 2017-11-22
Some ClayDropdown fixes

carloslancha:


matuzalemsteles:

jbalsas:
Hey @matuzalemsteles, did you see anything weird here? I'm going to merge so we don't clash in other things, but go ahead with the review and let @carloslancha know of anything you find or questions you have so we can discuss it!

packages/clay-dropdown/src/ClayDropdown.js
```diff
@@ -41,6 +41,15 @@ ClayDropdown.STATE = {
...
 	expanded: Config.bool().value(false),
 
+	/**
+	 * Icon to use in the trigger.
+	 * @instance
+	 * @memberof ClayDropdown
+	 * @type {?string}
+	 * @default undefined
+	 */
+	icon: Config.string(),
```
matuzalemsteles, line 51: Why no using `symbol` and not `icon`?

Improve the API by passing `spritemap` into the `icon`. We generate a better consistency with the other components.
```javascript
icon: Config.shapeOf({
    spritemap: Config.string().required(),
    symbol: Config.string().value('caret-bottom')
}).required(),
```

Soy using:
```soy
{if $icon.symbol != 'caret-bottom'}
	{call ClayIcon.render}
	    {param spritemap: $icon.spritemap /}
            {param symbol: $icon.symbol /}
	{/call}
{/if}

{if $label}
	{$label}
{/if}

{if $icon.symbol == 'caret-bottom'}
	{call ClayIcon.render}
		{param spritemap: $icon.spritemap /}
		{param symbol: $icon.symbol /}
	{/call}
{/if}
```

Would it be less confusing what do you think?

matuzalemsteles: For some reason, the `symbol` is not set to the default value when it is not passed.

jbalsas: Hey @matuzalemsteles, this attempts to achieve exactly that. Check https://github.com/brianchandotcom/liferay-portal/pull/53237 to see how the components are used in the end. 

Having complex state hierarchies makes them way too difficult to use outside our simple scenario. We need to aim for more flat states and also remove prefixes as much as possible so the APIs are simpler. For instance, a dropdown will always show a `label` and an `icon`, no need to call that `triggerLabel` and `triggerIcon.symbol`.

matuzalemsteles: Yeah, It's very more complexity. 😅

PR #172, opened 2017-11-24
Add elementClasses and id and some fixes

carloslancha:


packages/clay-dropdown/src/ClayDropdownBase.soy
```diff
@@ -10,21 +10,27 @@
...
 	{@param? style: string}
 	{@param? triggerClasses: string}
 	{@param? type: string}
 
-	{let $wrapperClasses kind="text"}
-		dropdown
-		{if $expanded}
-			{sp}show
+	{let $attributes kind="attributes"}
+		classes="dropdown
```
matuzalemsteles: This should not be `class="`?

matuzalemsteles (changes requested):

packages/clay-dropdown/src/ClayDropdownBase.soy
carloslancha: Sure, I corrected it later when I realized that 😅

matuzalemsteles (approved):

matuzalemsteles:
Merged, Thanks!

PR #173, opened 2017-11-25
IFI-151 Show coverage the tests

matuzalemsteles:
### Disclaimer
Just changing so that jest shows the coverage of the tests.

### Coveralls
I think it's interesting to be able to show the coverage status of our tests and to follow through an interface as easy access.

[coveralls](https://coveralls.io/features) is a tool that can help us when we send some PR that add some functionality and we do not add tests and our coverage drops, the tool always sends a message showing how much pr impacts on coverage. In addition to other interesting features.
image

What do you think about this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "precommit": "lint-staged",
     "prettier": "prettier-eslint packages/clay-*/src/*.js packages/clay-*/src/**/*.js",
     "start": "http-server . -p 4000",
-    "test": "lerna run soy && jest && npm run a11y"
+    "test": "lerna run soy && jest --coverage && npm run a11y"
   },
   "devDependencies": {
     "babel-preset-metal": "^4.1.0",


carloslancha:
I prefer to use the jest configuration object over adding params to the command (`--coverage`). In this case we can generate and save the reports by adding this to the jest object in the `package.json`

```
  "jest": {
    "collectCoverage": true,
    "coverageReporters": ["json", "html", "lcov", "text"],
    "coverageDirectory": "./coverage", //this is the default
    ...
  },
```

Probably `lcov` is enough to make coveralls or any other tool work, but `html` produces a pretty intuitive coverage pages locally, highlighting the code you're not covering and why.

What do you think? 😄

matuzalemsteles:
Hey @carloslancha,

By using the `--coverage` flag, by default jest generates reports on `lcov` and `html` pages so that we can view 😅. We can leave these settings explicitly as well.

Just a suggestion of use to keep control of the pr related to coverage and also as a way to expose it to other people.

carloslancha:
Mmmm, I'm not getting the `html` generated pages by default @matuzalemsteles. Looks like the default values are `["json", "lcov", "text"]` (text is console output) https://facebook.github.io/jest/docs/en/configuration.html#coveragereporters-array-string

Please, confirm this, maybe I'm doing something wrong.

But yeah, the idea is the same, my point was about my preference on keeping this kind of things configured by json 😉

matuzalemsteles:
Are the files generated inside the `coverage/lcov-report` folder, can you see it?

carloslancha:
Ok, don't know why that folder was ignored in my editor config 😅  Thx!

carloslancha:
Hey @matuzalemsteles I was curious about coveralls and I've make a test to integrate it (seems sooo easy). You can take a look in https://github.com/metal/metal-clay-components/pull/176

jbalsas:
Closing in favour of #176. @matuzalemsteles, please review and merge if you're fine with it!

PR #176, opened 2017-11-27
Coverage

carloslancha:


package.json
```diff
@@ -3,8 +3,9 @@
   "scripts": {
     "a11y": "metal-a11y --packages ./packages -c demos/index.html",
     "build": "lerna run build",
-    "ci": "npm run build && npm run checkFormat && npm run lint && npm run test",
+    "ci": "npm run build && npm run checkFormat && npm run lint && npm run test && npm run coveralls",
```
matuzalemsteles: What do you think the using CI configurations in `travis.yml`?

```yml
script:
  - npm run build
  - npm run checkFormat
  - npm run lint
  - npm run test

# Coveralls
after_success: npm run coveralls
```

matuzalemsteles:
Hey @carloslancha, Coveralls exist a feature that fails in pr when coverage is reduced. Are we using this?

package.json
carloslancha: Sounds good!

carloslancha:
Hey @matuzalemsteles nope, for that I think we need to get on an agreement on how that should work, if should fail when coverage is reduced, if is reduced down a defined threshold... What do you think?

jbalsas:
If we enable it, we should add the coverage badge to the `README.md` file as well 😉

matuzalemsteles:
@carloslancha Yeah, it has the option of when a PR reduces **x%** coverage of the repo it fails and we can also put the minimum coverage. I do not have the right margin to measure this. I think we can go without it for now and when we realize that it is not having a good coverage we can measure this percentage better. Do you have any suggestions about this?

carloslancha:
Sure thing @jbalsas I was waiting for the first report to get the url ;)

carloslancha:
@matuzalemsteles Yeah, I think we can adjust it later, maybe after talking with te rest of the team :)

matuzalemsteles:
@carloslancha Okay. 

Merged, thanks!

PR #177, opened 2017-11-27
Update browserslist-config on the generator

matuzalemsteles:


carloslancha:
Hey @matuzalemsteles we're using alpha.3 in the rest of the components, shouldn't be here the same version?

matuzalemsteles:
Hey @carloslancha, Yes, but it seems that the current version of `browserslist-config-clay-components` is `alpha.2` I believe it is not accompanying the versions of the components because they do not have a change.

jbalsas:
Versions are automatically released when the package has changes at the time of publication, yes 👍

carloslancha:
Merged, thx!

PR #185, opened 2017-11-29
[DO NOT MERGE] Merge Clay into Clay-Components

jbalsas:
Hey guys, this is an initial step to merge together https://github.com/liferay/clay and https://github.com/metal/metal-clay-components into one.

Since this is already a lerna repo, it's impossible to directly import it into `clay` directly, so I took the other way around, meaning that we are here importing every commit from that repo and putting it into `packages/clay`. After this is completed, we will probably need forcibly replace all history in https://github.com/liferay/clay with this one...

The cool thing now is that changes in `packages/clay` are automatically available locally linked for any other component in this repo. We can start collaborating closer and analyzing how different changes might affect our components. We can also start considering what type of distribution mechanism we want (claycss.zip with css and components...). Finally, we can do synced releases as a whole.

I'm reaching out to @pat270, @Robert-Frampton, @carloslancha, @matuzalemsteles and @marcoscv to provide some input, since this might impact your workflows for a little while :S


jbalsas:
(I see I might need to rebase on my own, but this is good to at least start the conversation)

jbalsas:
@pat270, can you list the usual tasks you used to perform so I can double check they work the same in this layout for now? Like, release, site update, publish, git flow...

pat270:
@jbalsas build, serve, release, toc

jbalsas:
Cool, I'll take a look and work out some of the lateral issues before continuing with this!

Issue #182, opened 2017-11-28
Power to customize ClayCheckbox and ClayRadio


matuzalemsteles:
Currently **ClayCheckbox** and **ClayRadio** can not customize your label, there are cases like `Card` that we can make a `Card` as a checkbox.

I think we can make ClayCheckbox with a `ClayCheckboxBase` base class so we can have more customization power and make the `ClayCheckbox` class available to use as the current one.

```
- ClayCheckbox.js
- ClayCheckboxBase.js
```

In this thought we keep Checkbox markup control centralized in one place `ClayCheckboxBase`.

What do you guys think about that?

carloslancha:
I think what you need can be done adding a parameter called, don't know, `content` and render it instead `<span class="{$spanLabelClasses}">{$label}</span>`

What do you think?

matuzalemsteles:
Hey @carloslancha, I did not want to change the existing markup of the Card, currently, when a card is a checkbox its markup looks like this:
```html
<div class="custom-control custom-checkbox">
    <label>
        <input class="custom-control-input" type="checkbox">
        <span class="custom-control-indicator"></span>
        <div class="card...">...</div>
    </label>
</div>
```

carloslancha:
I know @matuzalemsteles basically there we're changing the `span` with the label text to a custom `div`, that's why I said about adding a `content` param for those cases we wan't to customize more the label.

matuzalemsteles:
Sorry @carloslancha, had not understood well 😅. This approach is also simpler, but we will give the freedom that others can customize it, is this what we want? What do you think about this?

carloslancha:
In this case I feel creating a `ClayCheckboxBase` or something like that is a little overkill for this and I'd go for a simpler approach even if that means more flexibility to the user. Both are good solutions, but I don't know...

@jbalsas any thoughts about this? 😛

jbalsas:
It is of course convenient to allow any kind of markup injection, and the more we do the more it looks like we might end up there, but I don't like the idea...

I'm not sure how everything is currently setup, but maybe we could do something like this:

We currently have:

```soy
{template .input}
        ...
	<label>
		{let $inputAttributes kind="attributes"}
		    ...
		{/let}

		<input {$inputAttributes}></input>

		<span class="custom-control-indicator"></span>

		{let $spanLabelClasses kind="text"}
			custom-control-description
			{if $hideLabel}
				{sp}sr-only
			{/if}
		{/let}

[...]

PR #186, opened 2017-11-30
Create checkbox label extension with deltemplates

carloslancha:
## Why this?

As @jbalsas proposed in https://github.com/metal/metal-clay-components/issues/182#issuecomment-348138332 the best way to extend a component to fit our concrete needs is with deltemplates.

## Why don't do it in this way?

This can be done by simply creating a deltemplate as an interface, with the params it's going to receive declared and then create variants in our modules, something like:

`ClayCheckbox.soy`
```soy
{template .render}
   {delcall ClayCheckbox.Label variant='{$labelVariant}'}
      {param label: $label /}
   {/delcall}
{/template}

{deltemplate ClayCheckbox.Label}
   {@param label: html|string}

   <span class="{$spanLabelClasses}">{$label}</span>
{/deltemplate}
```

`ClayCard.soy`
```soy
{template .render}
...
   {call Checkbox.render}
      {param label: 'myLabel' /}
      {param labelVariant: 'card' /}
   {/call}
...
{/template}

{deltemplate ClayCheckbox.Label variant="'card'"}
   {@param label: html|string}

   <div class="card">
      .... {$label} ....
   </div>
{/deltemplate}
```

But with this approach we are limited by the contract with the default implementation of the `ClayCheckbox.Label`, I mean, if we want to pass it a parameter `coolParameter` we only can do it by adding it (obviously as optional) in the default implementation and all over the rest of implementations to comply the contract.

## Why to do it in this way?

So I prefer the following approach:

`ClayCheckbox.soy`
```soy
{template .render}
   {delcall ClayCheckbox.Label variant='{$labelVariant}' data="all" /}
{/template}

{deltemplate ClayCheckbox.Label}
   {call .label data="all" /}
{/deltemplate}

{template .label}
   {@param label: html|string}

   <span class="{$spanLabelClasses}">{$label}</span>
{/template}
```

`ClayCard.soy`
```soy
{template .render}
...
   {call Checkbox.render}
      {param coolParameter: 'Wow it's so cool!' /}
      {param label: 'myLabel' /}
      {param labelVariant: 'card' /}
   {/call}
...
{/template}

{deltemplate ClayCheckbox.Label variant="'card'"}
   {call .cardLabel data="all" /}
{/deltemplate}

{template .cardLabel}
   {@param coolParameter: string}
   {@param label: html|string}

   <div class="card">
      {$coolParameter}
      .... {$label} ....
   </div>
{/template}
```

As you can see here we do not have to define everywhere the `coolParameter` but where it's being used, in this case in the `.cardLabel` template inside my `ClayCard` component, and we don't need to worry about the rest of extension that could exists.

### DIFF
--- a/packages/clay-checkbox/src/ClayCheckbox.js
+++ b/packages/clay-checkbox/src/ClayCheckbox.js
@@ -96,6 +96,15 @@ ClayCheckbox.STATE = {
 	label: Config.any().required(),
 
 	/**
+	 * Name of the label template variant.
+	 * @instance
+	 * @memberof ClayCheckbox
+	 * @type {?string}
+	 * @default undefined
+	 */
+	labelVariant: Config.string(),
+
+	/**
 	 * Name to be applied to the element.
 	 * @instance
 	 * @memberof ClayCheckbox
--- a/packages/clay-checkbox/src/ClayCheckbox.soy
+++ b/packages/clay-checkbox/src/ClayCheckbox.soy
@@ -12,6 +12,7 @@
 	{@param? id: string}
 	{@param? indeterminate: bool}
 	{@param? inline: bool}
+	{@param? labelVariant: string}
 	{@param? name: string}
 	{@param? value: string}
 
@@ -37,6 +38,7 @@
 			{param disabled: $disabled /}
 			{param hideLabel: $hideLabel /}
 			{param label: $label /}
+			{param labelVariant: $labelVariant /}
 			{param name: $name /}
 			{param value: $value /}
 		{/call}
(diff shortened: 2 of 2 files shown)


matuzalemsteles:
Hey @carloslancha, this is cool. But I have the impression that we have the same thing with its proposed approach to using `content`, In soy some devs could also use this approach we are doing for our needs, as we do not want to use `deltemplates` or `data="all"`, perhaps the content approach would be more viable? Or do you find this approach a more elegant way to do it?

carloslancha:
Yeah @matuzalemsteles it was a proof of concept. After talking with @jbalsas we're going with the `$labelContent` approach.

Thx!

PR #191, opened 2017-11-30
Adds 'colors' to STATE config

bryceosterhaus:
State config for `colors` was missing. It is used at https://github.com/metal/metal-clay-components/blob/master/packages/clay-charts/src/ChartBase.js#L206


robframpton (approved):

carloslancha:
Merged, thx!

PR #194, opened 2017-12-03
 Add labelContent support for ClayRadio

matuzalemsteles:
Just adding `labelContent` to ClayRadio to do the same as ClayCheckbox.


carloslancha:
Merged, thx!

Issue #144, opened 2017-11-13
Passing parameters `data` to ClayLink


matuzalemsteles:
In JSP some users need to pass attributes to their components and one way was to use the `data` parameter like **HashMap**, Use cases like when you want a link not to use the SPA, passing `data-senna-off`.

carloslancha:
We need to define a convention for this. My suggestion would be to make components accept a param named "dataAttributes" containing a map.

```
{call Component.render}
   {param dataAttributes: [
      'id': '23',
      'myAttribute': 'myValue',
      ....
   ] /}
   ...
{/call}
```

and in the component use it just like `elementClasses`:

```
{let $attributes kind="attributes"}
   {if $dataAttributes}
      {foreach $attributeKey in keys($dataAttributes)}
         data-{$attributeKey}="{$dataAttributes[$attributeKey]}"
      {/foreach}
   {/if}

   {if $elementClasses}
      {$elementClasses}
   {/if}

   ...
{/let}

matuzalemsteles:
Hey @carloslancha, Sounds good to me, do you think we should use in more other components?

carloslancha:
@matuzalemsteles maybe is a good idea to add `elementClasses`, `id` and `dataAttributes` to all components by default and add it as a good practice to the guidelines.

matuzalemsteles:
@carloslancha I like this idea, I think we can do it.

matuzalemsteles:
Hey @carloslancha, The `SoyToIncrementalDomSrcCompiler` does not support using variables to transform into attributes. There is a question about this https://github.com/google/closure-templates/issues/112.

carloslancha:
![facepalm](https://media2.giphy.com/media/tZiLOffTNGoak/giphy.gif)

carloslancha:
Hey @matuzalemsteles 

@jbalsas had an idea about how to handle this, I'm working on it!

PR #197, opened 2017-12-05
Pass parameter 'data'

carloslancha:
Once component is attached all data attributes passed are bounded to the element and can be retrieved with metal domData:

`domData.get(component.element, 'dataAttributeName')`

You can also retrieve it from the element received in an event, for example:

```soy
{call ClayLink.render}
   {param data: ['myAttr': 'myValue'] /}
   {param events: ['click': $myHandler_] /}
   {param label: 'My Link' /}
{/call}
```
```js
myHandler_(event) {
   console.log(domData.get(event.target, 'myAttr')); //Console output would be 'myValue'
}
```

This solves https://github.com/metal/metal-clay-components/issues/144

### DIFF
--- a/packages/clay-button/src/ClayButton.js
+++ b/packages/clay-button/src/ClayButton.js
@@ -16,7 +16,9 @@ class ClayButton extends Component {
 	 */
 	attached() {
 		for (let dataKey in this.data) {
-			domData.get(this.element, dataKey, this.data[dataKey]);
+			if (this.data.hasOwnProperty(dataKey)) {
+				domData.get(this.element, dataKey, this.data[dataKey]);
+			}
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


matuzalemsteles:

matuzalemsteles:
@carloslancha a great solution 🎉. But just to clarify a point, since usually use `senna-off` quite explicitly in the element, may have to add support for it to senna. What do you think about this?

carloslancha:
@matuzalemsteles what do you mean with "may have to add support for it to senna"?

matuzalemsteles:
@carloslancha Why does senna observe element selectors, since `domData` only adds in a `__METAL_DATA__` object, **senna** could not see for example. In the case of using the `senna-off` attribute on ClayLink.

carloslancha:
Working on it @matuzalemsteles I think we've find a solution!

matuzalemsteles:
Merged, thanks!

PR #200, opened 2017-12-05
Improve ClayComponent tests

### LINKED ISSUE #144: Passing parameters `data` to ClayLink

carloslancha:
This resolves https://github.com/metal/metal-clay-components/issues/144


carloslancha:
@matuzalemsteles this should resolve what you said about data-senna-off

matuzalemsteles:
Hey @carloslancha, I like the idea of having a component to centralize common behaviors. What do you think of this package if call `shared` and not make this package public, it is necessary to make it available? we could add things in common that we have between more than one package, something like [facebook uses react](https://github.com/facebook/react/tree/master/packages/shared). Just a thought to know if this fits into our reality.

jbalsas:
Hey @matuzalemsteles, we probably want to expose `ClayComponent` as consumers will probably want to build on top of that if they are using Clay.

I'll merge ahead so I can do a quick release in sync with the latest clay one, but keep the discussion going, figure out what else could be abstracted away here and apply to the rest of the components whenever you can!

PR #202, opened 2017-12-06
Updates clay to v2.0.0-beta4

jbalsas:


jbalsas:
Hey @matuzalemsteles, @carloslancha, I've quarantined the `metal-clay-generator:app` test because it is failing on Travis more often than not. We need to take a look to figure out why :)

PR #217, opened 2017-12-15
Create ClayTable

matuzalemsteles:
### Disclaimer
This is an implementation of the lexicon [table](https://lexicondesign.io/docs/patterns/Table/table.html) pattern, I am not sure if we will implement with this component the definition of [inline edit table](https://lexicondesign.io/docs/patterns/Table/inlineEditTable.html) or we will only separate into another component and if it will be more feasible for the experience of using the component.

> Test with clay#develop for use `table-cell-expand`

### Missing features
* Columns drag and drop
* Sort columns

### API
The API tries to follow the definition of Lexicon, how to use just the `Sticker` before the first column, Also having to leave the custom table to supply all the use cases like adding `colspan` to a `th` and adding custom classes in `td` and `th`.

I'm not sure about the `useEllipse` API name that adds an ellipse when the text overflows its size.

### Columns drag and drop
I've been talking to @pat270 in the last few days about how d&d would be implemented in columns of the tables, Patrick even prepared a [visual test](http://pat270.github.io/lexicon/v2.0.0-table-drag-poc/content/test-table-drag-column/) and how to proceed, but we've been wondering if this is performed and will generate many bugs, so we decided to work with this functionality more cautiously and implement it in the best possible way.


carloslancha:

carloslancha:

carloslancha:
Resent https://github.com/metal/metal-clay-components/pull/222

PR #218, opened 2017-12-19
Update clay-charts to billboard 1.2.0

julien:


carloslancha:
LGTM

packages/clay-charts/src/ChartBase.js
```diff
@@ -24,6 +24,30 @@ const DEFAULT_COLORS = [
...
+	'bb-line-dashed-2-4',
+	'bb-line-dashed-3-2',
+	'bb-line-dashed-3-3',
+	'bb-line-dashed-3-4',
+	'bb-line-dashed-4-2',
+	'bb-line-dashed-4-3',
+	'bb-line-dashed-4-4',
+];
+
+const DEFAULT_PATTERNS = [
```
jbalsas: This API is actually `tiles`. Why don't we rename all of this so the relation is more obvious?

julien: @jbalsas yes, it makes sense. I renamed patterns to tiles.

PR #219, opened 2017-12-20
Rename pattern to tiles and add default point patterns

julien:


packages/clay-charts/package.json
```diff
@@ -18,9 +18,10 @@
   "scripts": {
     "build": "npm run soy && webpack",
     "compile": "babel -d lib/ src/ -s --ignore src/__tests__",
-    "prepublish": "npm run soy && npm run sass && npm run compile",
+    "prepublish": "npm run soy && npm run sass && npm run compile && npm run copy",
     "sass": "node-sass src/scss/main.scss -o lib/css --source-map true",
-    "soy": "metalsoy --soyDeps node_modules/metal*/src/**/*.soy node_modules/clay*/src/**/*.soy"
+    "soy": "metalsoy --soyDeps node_modules/metal*/src/**/*.soy node_modules/clay*/src/**/*.soy",
+    "copy": "ncp src/svg lib/svg"
```
jbalsas: let's call this maybe `copyAssets`? or `copySvg`? :)

packages/clay-charts/src/svg/tiles.svg
```diff
@@ -1,55 +1,46 @@
 <svg xmlns="http://www.w3.org/2000/svg">
```
jbalsas: Which would be the way to insert this in the DOM in an inert way? (not showing)

julien: @jbalsas for the moment we can use `<svg ... width="0" height="0">`  or `<svg ... style="display:none">` until we find the "idiomatic" way of doing it.

PR #234, opened 2017-12-26
Update markups and fix

matuzalemsteles:
hey @carloslancha, `ClayUserCard` customize the `label` of `ClaySticker` mean that we are duplicating the markup, but it is obliged to use the `sticker-overlay` class in the label to be positioned correctly, we can add as default in `ClaySticker` but I do not know what effects it can cause it being used on other components. What do you think about this?

> This pr was tested with #233.

### DIFF
--- a/packages/clay-sticker/src/ClaySticker.soy
+++ b/packages/clay-sticker/src/ClaySticker.soy
@@ -72,7 +72,21 @@
 	{@param? imageAlt: string}
 	{@param? imageSrc: string}
 
-	<img alt="{$imageAlt}" class="img-fluid" src="{$imageSrc}" />
+	{let $attributes kind="attributes"}
+		class="sticker-img"
+
+		{if $imageAlt}
+			alt="{$imageAlt}"
+		{/if}
+
+		{if $imageSrc}
+			src="{$imageSrc}"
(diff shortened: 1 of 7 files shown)


packages/clay-sticker/src/ClaySticker.soy
```diff
@@ -72,7 +72,21 @@
 	{@param? imageAlt: string}
 	{@param? imageSrc: string}
```
carloslancha, line 73: In fact this two params should be mandatory for this `.image` template, so we can get rid of `let` and just place the variables in the image, or at least remove the conditions.

carloslancha:

carloslancha (changes requested):
Since class `.sticker` has `position: relative` adding a `span` inside with `.sticker-overlay` should not be a problem. Go for it, add it as default wrapper on all stickers, remove it from `ClayUserCard` and update the ussage by passing `imageSrc` to it. Nice catch @matuzalemsteles !

Btw... stop working and enjoy your holidays!! 😂

carloslancha:

carloslancha:
Resent https://github.com/metal/metal-clay-components/pull/237

PR #235, opened 2017-12-26
EnableMcritic

### LINKED ISSUE #204: Enable fully mcritic testing
We are currently ignoring a lot of packages from soy linting because it gives us problems. We need to enable it again or we risk introducing bugs and inconsistencies.

carloslancha:
This fixes https://github.com/metal/metal-clay-components/issues/204

### DIFF
--- a/packages/clay-card/src/ClayCard.js
+++ b/packages/clay-card/src/ClayCard.js
@@ -1,12 +1,12 @@
 import 'clay-checkbox';
-import 'clay-dropdown';
 import 'clay-label';
 import 'clay-link';
 import 'clay-sticker';
+import {ClayActionsDropdown} from 'clay-dropdown';
+import {Config} from 'metal-state';
 import Component from 'metal-component';
 import defineWebComponent from 'metal-web-component';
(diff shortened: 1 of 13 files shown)


jbalsas (dismissed):
Hey @carloslancha, I'm not sure these are the right changes, we should have different regex settings so we can target the current code without the need of rewriting the imports as named imports and disabling eslint for them...

packages/clay-card/src/ClayCard.js
```diff
@@ -1,12 +1,15 @@
+/* eslint-disable */
+import { ClayActionsDropdown } from 'clay-dropdown';
+/* eslint-disable */
```
jbalsas: This should probably be enable 😉

packages/clay-card/src/ClayHorizontalCard.js
```diff
@@ -1,11 +1,14 @@
+/* eslint-disable */
+import { ClayActionsDropdown } from 'clay-dropdown';
+/* eslint-disable */
```
jbalsas: disable 😉

PR #254, opened 2018-01-10
Update ClayTable, ClayList and ClayCardGrid schemas. Add plus button as a dropdown feature to Management Toolbar

carloslancha:


matuzalemsteles:

packages/clay-card-grid/demos/index.html
```diff
@@ -27,198 +27,133 @@ <h1 class="page-title">
...
 			{
 				href: '#1',
-				label: 'Option 1',
+				label: 'Remove',
 			},
 			{
 				href: '#2',
-				label: 'Option 2',
+				label: 'Download',
 				separator: true,
```
matuzalemsteles, line 38: Just remove the `separator` since it is not necessary anymore in the demo. 🙂

matuzalemsteles:

packages/clay-card-grid/src/ClayCardGrid.soy
```diff
@@ -49,29 +53,33 @@
  */
 {template .item}
```
matuzalemsteles, line 54: Is this template being used?

matuzalemsteles (commented):
hey @carloslancha, realizes that when we open the dropdown for the first time the `separator` is rendered and after we close and open again it disappears, this happens with more than 1 `separator` and only when not using groups.

Notice that the `separator` value comes `undefined` in ClayDropdownItem not respecting its default value when it is not passed.

> I'll have to update the examples again at claycss.com 😅

q1f2gjljno

packages/clay-card-grid/demos/index.html
carloslancha: I've added another element so now the separator makes sense.

packages/clay-card-grid/src/ClayCardGrid.soy
carloslancha: I've refactored all this stuff

carloslancha:
@matuzalemsteles nice catch! 

FYI: It was a bug caused by the creation of `ClayDropdownItem` component. I didn't realize that component was rendering, in some cases like in the separator true one, more than one element, and each complete component (the output of the .render template) can only render one element, all nodes after that are ignored.

PR #255, opened 2018-01-11
Update webpack config to use sourcemaps

julien:


carloslancha:
Nice @julien ! Thx, merged!

PR #256, opened 2018-01-11
ClayDatasetDisplay

carloslancha:


matuzalemsteles:

packages/clay-dataset-display/src/ClayDatasetDisplay.js
```diff
@@ -0,0 +1,472 @@
...
+	icon: Config.string(),
+	label: Config.string().required(),
+	quickAction: Config.bool(),
+	separator: Config.bool().value(false),
+	type: Config.oneOf(['group', 'item']).value('item'),
+};
+
+const actionItemsValidator = Config.arrayOf(Config.shapeOf(actionItemShape));
+
+actionItemShape.items = actionItemsValidator;
```
matuzalemsteles: What do you think of putting that block above validators in a separate file?

packages/clay-dataset-display/src/ClayDatasetDisplay.js
```diff
@@ -0,0 +1,472 @@
...
+	 * @type {?string|undefined}
+	 * @default undefined
+	 */
+	searchInputName: Config.string(),
+
+	/**
+	 * Flag to indicate if the dataset is selectable.
+	 * @instance
+	 * @memberof ClayDatasetDisplay
+	 * @type {?bool|undefined}
```
matuzalemsteles: SF: Should not be `{?bool}`?

packages/clay-dataset-display/src/ClayDatasetDisplay.js
```diff
@@ -0,0 +1,472 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Iterates over the selected items array, mark all as not selected and
+	 * removes them from the selectedItems array.
+	 * @private
+	 */
+	deselectAllItems() {
```
matuzalemsteles: Would not it be better to follow the default `deselectAllItems_`? Or was it intentional?

matuzalemsteles (commented):
Great working on the component 🎉!

packages/clay-dataset-display/src/ClayDatasetDisplay.js
carloslancha: In fact since last changes in mcritic we can share validators between components and latest changes in metal-state Config should not mutate items anymore so... I'm going to give it a try.

packages/clay-dataset-display/src/ClayDatasetDisplay.js
carloslancha: D'oh!

packages/clay-dataset-display/demos/index.html
```diff
@@ -0,0 +1,418 @@
...
+							inputValue: '3',
+						},
+					],
+					inputName: 'item1radio',
+					label: 'Group 1',
+					type: 'radiogroup',
+				}
+			],
+			items: items,
+			plusButton: {
```
jbalsas, line 251: What does this API refer to in particular? The Management Toolbar add actions? We should find a better name for this item...

jbalsas:
I'm merging ahead, please think about the `plusButton` API and send a followup PR!

packages/clay-dataset-display/demos/index.html
carloslancha: https://github.com/metal/metal-clay-components/pull/259

PR #258, opened 2018-01-12
Re-enable a11y validation and fix issues

carloslancha:


matuzalemsteles:

README.md
```diff
@@ -43,3 +43,6 @@
   ```
   npm run test
   ```
+
+## Important
+Since travis is failing on executing a11y tests due to out of date suid you need to execute `npm run a11y` locally before sending any pr to verify we're following accessibility standars.
```
matuzalemsteles, line 48: What do you think of adding `npm run a11y` in `lint-staged`?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "a11y": "metal-a11y --packages ./packages -c demos/index.html",
+    "a11y": "metal-a11y --packages ./packages -c demos/a11y.html",
```
matuzalemsteles, line 4: hey @carloslancha, why two files in the `index` and `a11y` demos?

carloslancha: `index` is for demo purposes, `ally` only for accessiblity. There're times when we are not following accessiblity due to the config we pass to the components (like in ClayList, where we're passing the same items to all instances in behalf of simplicity) but that doesn't mean component itself is not following a11y. That's where `ally.html` shows up.

README.md
carloslancha: I think `lint-staged` is fine for linting and format stuff, but if we add `ally` we force each commit yo be accessible, which I think is not a good idea while developing.

matuzalemsteles: Yeah, that might have gone bad, I had not thought about that part 😅.

package.json
matuzalemsteles: This is good. The bad thing is that we will have another file for maintenance.

carloslancha: When we have all Clay+claycss.com+components fully linked in the project demos probably will be useless. Since then...

PR #263, opened 2018-01-13
Create ClayPagination and ClayPaginationBar

carloslancha:
captura de pantalla 2018-01-13 a las 18 01 57

### DIFF
--- a/packages/clay-dropdown/src/ClayDropdownBase.js
+++ b/packages/clay-dropdown/src/ClayDropdownBase.js
@@ -209,6 +209,15 @@ ClayDropdownBase.STATE = {
 	caption: Config.string(),
 
 	/**
+	 * Content Renderer name of items.
+	 * @instance
+	 * @memberof ClayDropdownBase
+	 * @type {?string|undefined}
+	 * @default undefined
+	 */
+	contentRenderer: Config.string(),
+
+	/**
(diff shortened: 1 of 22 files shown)


carloslancha:
Hey @matuzalemsteles! Here I've tried to make all of the rendering in soy to make possible server-side rendering. Take a look and tell me your thoughts;)

jbalsas:
Hey @matuzalemsteles, I'm going to merge this because I want to push forward with the clay reunification, so this is likely to be the last PR in this repo before we move everything to http://github.com/liferay/clay

matuzalemsteles:
hey @carloslancha, Good job with rendering on SSR. But just pointing out a few points.

* When we reach more than 6 pages and `currentPage` is the last page, the range ends up adding more than it should.
screen shot 2018-01-15 at 10 21 33

* We may have to trigger some events when there are clicks on the items.
* Will we let SPA always load pages when there is user interaction or are we going to give the possibility that devs will load the data when events are triggered? since it can gain some "performance" ... Using the events triggered by pagination.

What do you think?

carloslancha:
Hey @matuzalemsteles I'll take a look on that bug, didn't notice, thx!

I've preferred for now not to add those events till we need it. We can easily add it in the future.

EDIT: Fixed here https://github.com/metal/metal-clay-components/pull/265

PR #265, opened 2018-01-15
Fix ClayPagination range when current page is last page

carloslancha:


jbalsas:
Hey @carloslancha, @matuzalemsteles, please hold on on merging this to see if I can complete the migration to github.com/liferay/clay so we can move there!

